PR #4, opened 2023-08-02
Fix/minor fix for init

HYLcool:
* refine links and code in docs
* replace with smaller images and put them all into the docs/imgs directory


yxdyc (approved):
LGTM

PR #6, opened 2023-08-02
fix the wrong links in demos/*/app.py

yxdyc:
As the title says. Please also double-check whether I have updated the wrong links in ModelScope.


yxdyc:
By the way, the modelscope demo ["diversity"](https://modelscope.cn/studios/Data-Juicer/data_visulization_diversity/files) has no demo-dataset now. Plz upload a suitable one later @zhijianma

HYLcool (approved):

PR #19, opened 2023-09-13
Feature/make cache work for DJ & cache compression

HYLcool:
- **Cache Management** now works for Data-Juicer. It failed before because the original cache management from hugging face cannot serialize some of the process methods.
- **Cache Compression** is available now for space-time trade-off. For very large datasets (over tens of GB), cache compression in parallel requires relatively little extra time but could save significant disk space for caches.
  - It's only enabled when both cache and cache compression are enabled.
  - Now we support 3 compression algorithms: ["lz4" (fastest), "zstd" (highest compression rate), "gzip"]


data_juicer/core/data.py
```diff
@@ -161,7 +166,72 @@ def map(self, *args, **kargs):
...
-        return NestedDataset(super().map(*args, **kargs))
+        if 'new_fingerprint' not in kargs or kargs['new_fingerprint'] is None:
+            new_fingerprint = generate_fingerprint(self, *args, **kargs)
+            kargs['new_fingerprint'] = new_fingerprint
+
+        if cache_utils.CACHE_COMPRESS:
+            decompress(self,
+                       kargs['new_fingerprint'],
+                       kargs['num_proc'] if 'num_proc' in kargs else 1)
+
```
zhijianma, line 177: 是否加入自动清理cache files，避免excutor和analyser中的显式调用dataset的`cleanup_cache_files`?

HYLcool: Good idea~

yxdyc (approved):
LGTM

PR #24, opened 2023-09-18
Add new OP: token_num_filter

HYLcool:
- Add a new OP: token_num_filter
  - Allow specify any valid tokenizer from hugging face to compute the token number of samples


zhijianma (approved):
LGTM

PR #35, opened 2023-10-16
Modify the location of RayExecutor import

chenhesen:
When using DJ to process large-scale data, there will be a ray bug.  
https://github.com/ray-project/ray/issues/17745
Currently, it is adjusted to import ray only when using it.


HYLcool (approved):
LGTM

PR #51, opened 2023-10-30
Feat/add chinese mapper

chenhesen:
add chinese_convert_mapper:  Mapper to convert Chinese between Traditional Chinese, Simplified Chinese and Japanese Kanji
add remove_non_chinese_character_mapper: Mapper to remove non chinese Character in text samples


configs/config_all.yaml
```diff
@@ -83,6 +83,9 @@ process:
   - sentence_split_mapper:                                  # split text to multiple sentences and join them with '\n'
       lang: 'en'                                              # split text in what language
   - whitespace_normalization_mapper:                        # normalize different kinds of whitespaces to English whitespace.
+  - chinese_convert_mapper:                                 # Mapper to convert Chinese between Traditional Chinese, Simplified Chinese and Japanese Kanji.
```
HYLcool, line 86: add new operators in alphabetical order for each type of operator

HYLcool:
New operators need to be added to the operator lists in docs/Operators.md and docs/Operators_ZH.md. Besides, the numbers of each type of operator need to be updated in these two docs as well.

HYLcool (changes requested):
New operators need to be added to the operator lists in docs/Operators.md and docs/Operators_ZH.md. Besides, the numbers of each type of operator need to be updated in these two docs as well.

HYLcool (approved):

PR #74, opened 2023-11-15
Feature/image shape filter

HYLcool:
- Add image_shape_filter to filter out samples that contain images with out-of-range widths or heights.
- Limit the version of pyarrow <= 13.0.0 to avoid useless FutureWarning
- Convert a reg pattern string to raw string to avoid W605 warning


zhijianma (approved):
LGTM

PR #100, opened 2023-11-28
Feature/image text matching

zhijianma:
1. add image_text_matching_filter
2. rename clip_similarity_filter to image_text_similarity_filter, and add flip option


data_juicer/ops/filter/image_text_similarity_filter.py
```diff
@@ -9,26 +9,30 @@
...
-class ClipSimilarityFilter(Filter):
+class ImageTextSimilarityFilter(Filter):
     """Filter to keep samples those similarity between image and text
     within a specific range."""
 
     def __init__(self,
                  hf_clip='openai/clip-vit-base-patch32',
                  min_score: ClosedUnitInterval = 0.1,
                  max_score: ClosedUnitInterval = 1.0,
+                 horizontal_flip: bool = True,
```
HYLcool, line 34: why is the default value of horizontal_flip set to True?

docs/Operators.md
```diff
@@ -77,11 +77,12 @@ All the specific operators are listed below, each featured with several capabili
...
-| clip_similarity_filter      | Multimodal |   -    |  Keeps samples with similarity between text and images within the specified range           |
 | flagged_words_filter           | General | en, zh | Keeps samples with flagged-word ratio below the specified threshold                        |
 | image_aspect_ratio_filter      | Image   |   -    | Keeps samples contains images with aspect ratios within specific range                     |
 | image_shape_filter             | Image   |   -    | Keeps samples contains images with widths and heights within specific ranges               |
-| image_size_filter      | Image   |   -    | Keeps samples contains images whose size in bytes are within specific range                     |
+| image_size_filter              | Image   |   -    | Keeps samples contains images whose size in bytes are within specific range                     |
+| image_text_matching_filter     | Multimodal |   -    |  Keeps samples with matching score between text and images within the specified range           |
+| image_text_similarity_filter   | Multimodal |   -    |  Keeps samples with similarity between text and images within the specified range           |
```
HYLcool, line 85: I think it's better to show that "matching score" is from blip and "similarity" is from "clip". Otherwise, they are too similar to tell them apart.

data_juicer/ops/filter/image_text_matching_filter.py
```diff
@@ -0,0 +1,178 @@
...
+from data_juicer.utils.constant import Fields, StatsKeys
+from data_juicer.utils.mm_utils import SpecialTokens, load_image
+from data_juicer.utils.model_utils import get_model, prepare_model
+
+from ..base_op import OPERATORS, Filter
+from ..op_fusion import LOADED_IMAGES
+
+OP_NAME = 'image_text_matching_filter'
+
+with AvailabilityChecking(['torch'], OP_NAME):
```
HYLcool, line 14: In the first argument of AvailabilityChecking, we also need to add `transformers` to the dependency list. Same in the image_text_similarity_filter OP.

data_juicer/ops/filter/image_text_matching_filter.py
```diff
@@ -0,0 +1,178 @@
...
+    import torch
+    import transformers  # noqa: F401
+    from PIL import ImageOps
```
HYLcool, line 17: PIL is required by `datasets`. After we install the minimal requirements that include `datasets`, the Pillow/PIL is installed already. So we don't need to check the availability of PIL and we can move this import statement before the availability checking.

Same in the image_text_similarity_filter OP.

data_juicer/ops/filter/image_text_matching_filter.py
```diff
@@ -0,0 +1,178 @@
...
+            key: value
+            for key, value in SpecialTokens.__dict__.items()
+            if not key.startswith('__')
+        }
+        offset = 0
+
+        def remove_special_token(text):
+            for value in special_token_dict.values():
+                text = text.replace(value, '')
+            return text
```
HYLcool, line 115: It's better to extract this function into the common module.

data_juicer/ops/filter/image_text_matching_filter.py
```diff
@@ -0,0 +1,178 @@
...
+            return True
+
+        keep_bools = np.array([
+            self.min_score <= itm_score <= self.max_score
+            for itm_score in itm_scores
+        ])
+
+        # different strategies
+        if self.any:
+
```
HYLcool, line 175: an extra blank line can be removed.

data_juicer/ops/filter/image_text_matching_filter.py
zhijianma: Done

data_juicer/ops/filter/image_text_matching_filter.py
zhijianma: Done

data_juicer/ops/filter/image_text_matching_filter.py
zhijianma: Done

PR #106, opened 2023-11-29
Better Analyzer support for stats of multimodal OPs & other minor enhancements

HYLcool:
- Add two arguments `keep_stats/hashes_in_res_ds` to control whether to keep stats (computed by Filters) or hashes (computed by Deduplicators) in the exported result dataset. In default, they are False.
  - From #102 
- Analyzer:
  - Save stats before doing the analysis
  - Support to apply overall analysis in parallel
  - Support to apply overall and column-wise analysis for DataFrame column of `list` type (some stats generated by multimodal OPs)


HYLcool:
+ bug fixed: unmatched OP name list and actual OP list after skipping some OPs whose dependencies are not installed. (From user Junqin Huang)

yxdyc (approved):
LGTM

PR #110, opened 2023-12-04
Feature/face_area_filter 

drcege:
1. add face_area_filter
2. use face detector from [dlib](http://dlib.net/imaging.html#get_frontal_face_detector)
3. compatible with multiprocess


data_juicer/ops/filter/face_area_filter.py
```diff
@@ -0,0 +1,133 @@
...
+        self.detector = dlib.get_frontal_face_detector()
+
+    def compute_stats(self, sample, context=False):
+        # check if it's computed already
+        if StatsKeys.face_ratios in sample[Fields.stats]:
+            return sample
+
+        # there is no image in this sample, still default ratio 0.0
+        if self.image_key not in sample or not sample[self.image_key]:
+            sample[Fields.stats][StatsKeys.face_ratios] = [0.0]
```
HYLcool, line 68: Setting it to 0 for samples without images might influence the overall distribution of face ratios. How about setting it to an empty float64 array like other OPs?
```
sample[Fields.stats][StatsKeys.image_text_matching_score] = np.array([], dtype=np.float64)
```

docs/.Operators_ZH.md.un~
HYLcool, line 1: Two extra useless files: 
- docs/.Operators_ZH.md.un~
- docs/Operators_ZH.md~

HYLcool (approved):
LGTM

PR #111, opened 2023-12-04
Two addition ops from DiHT

BeachWang:
1. text_action_filter: Filter to keep texts those contain actions in the text.
2. text_object_dependency_filter: Identify the objects in the text which are independent with other token, and filter them.
    The text containing no objects will be omitted.


tests/ops/data/.ipynb_checkpoints/cat-checkpoint.jpg
drcege, line 1: should remove caches

PR #122, opened 2023-12-06
Two addition filter for image captions (resubmit)

BeachWang:
Add descriptions in the documents for the two filter. Following is the two filters:
1. text_action_filter: Filter to keep texts those contain actions in the text.
2. text_entity_dependency_filter: Identify the entities in the text which are independent with other token, and filter them.
The text containing no entities will be omitted.


data_juicer/ops/filter/text_action_filter.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self,
+                 lang: str = 'en',
+                 min_action_num: int = 1,
+                 *args,
+                 **kwargs):
+        """
+        Initialization method.
+
+        :param lang: language of the text in the samples. 'en' for detection of
+            actions in English an'zh' for detection of actions in Chinese.
```
HYLcool, line 25: an'zh' --> and 'zh'

same in text_entity_dependency_filter

HYLcool (approved):
Others LGTM

PR #137, opened 2023-12-14
fix: 1. change relpath to abspath of dataset

zhijianma:
1. change relpath to abspath of dataset， because relpath is not supported by ray to load and save dataset.
2. In python3.8 on Mac system,  torch set_num_threads does not  work.


HYLcool (approved):
LGTM.

PR #139, opened 2023-12-18
Feature/phrase grounding recall filter

HYLcool:
- add a new OP: phrase_grounding_recall_filter
- unittest opt: remove hf models after some tests done to save the disk space
- llava_to_dj and dj_to_llava tools: support to keep only the captions of each QA sample


chenhesen:
Great

chenhesen (approved):

Issue #136, opened 2023-12-14
[Feat] Add a tool to process data automatically using the 3-sigma principle


yxdyc:
### Search before continuing 先搜索，再继续

- [X] I have searched the Data-Juicer issues and found no similar feature requests. 我已经搜索了 Data-Juicer 的 issue 列表但是没有发现类似的功能需求。

### Description 描述

Different datasets exhibit unique data distributions, necessitating tailored hyperparameters for optimal processing. A straightforward strategy is to apply Filter OPs that retain only the samples falling within the 3-sigma ranges for the OP's stats. We can offer a tool to facilitate this approach.

### Use case 使用场景

Given an initial recipe, the tool can automatically 
1.  analyze the dataset first
2. generate a refined recipe based on the distribution stats and 3-sigma principle
3. process the dataset using the new recipe

### Are you willing to submit a PR for this feature? 您是否乐意为此功能提交一个 PR？

- [X] Yes I'd like to help by submitting a PR! 是的！我愿意提供帮助并提交一个PR！

PR #140, opened 2023-12-18
add an auto-hpo tool for data-recipes based on 3-sigma rule

yxdyc:
- implemented an auto-hpo tool for data-recipes based on 3-sigma rule, #136 ,
a log example is as follow:
image

- added a helper class to build the mapping from OPs' op_name to their used stats_key
- added a config exporter


yxdyc:
some information about unit-test: 

passed in local machine (dlib==19.24.2), but failed in github
image

image

data_juicer/config/config.py
```diff
@@ -478,6 +490,37 @@ def display_config(cfg):
...
+        save the config object, some params are from jsonargparse
+    :param cfg: cfg object to save (Namespace type)
+    :param path: the save path
+    :param format: 'yaml', 'json', 'json_indented', 'parser_mode'
+    :param skip_none: Whether to exclude entries whose value is None.
+    :param skip_check: Whether to skip parser checking.
+    :param overwrite: Whether to overwrite existing files.
+    :param multifile: Whether to save multiple config files
+        by using the __path__ metas.
+    :param branch:
```
HYLcool, line 505: Need to add description to this arg. BTW, it seems that save() function doesn't have an argument of "branch", maybe recheck is needed.

tools/hpo/README.md
```diff
@@ -1,6 +1,23 @@
...
+$$P(|x-\mu| > 3\sigma) \leq 0.003$$
+
+To automate this process, we provide the tool which can be used as follows:
+```shell
+# cd tools/hpo
+python execute_hpo_3sigma.py --config <data-process-cfg-file-path>
+
+#e.g.,
+python execute_hpo_3sigma.py --config configs/process.yaml
+```
```
HYLcool, line 18: The tool usage code might need to be aligned with the ZH version.

tools/hpo/execute_hpo_3sigma.py
```diff
@@ -0,0 +1,72 @@
...
+                        new_val = stats_key_to_mean[stats_key] - \
+                                  3 * stats_key_to_std[stats_key]
+                    if 'max' in arg_name:
+                        new_val = stats_key_to_mean[stats_key] + \
+                                  3 * stats_key_to_std[stats_key]
+                    if new_val is not None:
+                        logger.info(f'Using 3-sigma rule, changed para '
+                                    f'{arg_name}={args[arg_name]} into '
+                                    f'{arg_name}={new_val}')
+                        args[arg_name] = new_val
```
HYLcool, line 56: There might be "nan" in mean and std for some string stats or other object stats (e.g. lang). Maybe a type check should be done before calculate the new_val.

HYLcool (approved):
LGTM

PR #143, opened 2023-12-19
Feature/replace content mapper

zhijianma:
1. Add  `replace content mapper`op to replace all content in text that matches a specific regular expression pattern with a designated replacement string.
2. Release some mapper initialization parameters, such as `pattern` and `repl`


HYLcool (approved):
LGTM

Issue #146, opened 2023-12-20
[Bug]: “The features can't be aligned because the key __dj_stats__ of ...” for line_length related OPs 


yxdyc:
### Before Reporting 报告之前

- [X] I have pulled the latest code of main branch to run again and the bug still existed. 我已经拉取了主分支上最新的代码，重新运行之后，问题仍不能解决。

- [X] I have read the [README](https://github.com/alibaba/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process. (Otherwise, we recommend that you can ask a question using the Question template) 我已经仔细阅读了 [README](https://github.com/alibaba/data-juicer/blob/main/README_ZH.md) 上的操作指引，并且在安装过程中没有错误发生。（否则，我们建议您使用Question模板向我们进行提问）

### Search before reporting 先搜索，再报告

- [X] I have searched the Data-Juicer [issues](https://github.com/alibaba/data-juicer/issues) and found no similar bugs. 我已经在 [issue列表](https://github.com/alibaba/data-juicer/issues) 中搜索但是没有发现类似的bug报告。

### OS 系统

Ubuntu

### Installation Method 安装方式

from source

### Data-Juicer Version Data-Juicer版本

v0.1.2

### Python Version Python版本

3.8

### Describe the bug 描述这个bug

When some length related OPs are used in Analyzer to process dataset containing zero-length samples, there can be un-aligned features, such as `'max_line_length': Value(dtype='int64', id=None` v.s. `'max_line_length': Value(dtype='float64'`

### To Reproduce 如何复现

for pile dataset, raw chunk 24, running

`export HF_DATASETS_CACHE=/mnt/data/.cache/huggingface/datasets
python /root/data-juicer/tools/hpo/execute_hpo_3sigma.py --config dj_refine_recipe_base.yaml --path_3sigma_recipe dj_refined_3_sigma/dj_refine_recipe_chunk_24.yaml --dataset_path raw/24.jsonl.zst --export_path dj_refined_3_sigma/refined_24.jsonl.zst.jsonl`

### Screenshots 截图

![image](https://github.com/alibaba/data-juicer/assets/67475544/264861cc-1736-4253-9d90-57d8fa42b10c)

PR #147, opened 2023-12-20
bugfix for un-aligned types (int v.s. float) caused by two OPs

yxdyc:
- bugfix for un-aligned types (int v.s. float) caused by two OPs, see #146 
- improve the usability for logfile_name when running several exps in the same workdir


data_juicer/ops/filter/average_line_length_filter.py
```diff
@@ -49,7 +49,7 @@ def compute_stats(self, sample, context=False):
                 sample[Fields.context][context_key] = lines
         sample[Fields.stats][StatsKeys.avg_line_length] = \
             len(sample[self.text_key]) / len(lines) \
-            if len(lines) != 0 else 0.0
+            if len(lines) != 0 else 0
```
HYLcool, line 52: average line length could be float.

HYLcool (approved):
LGTM

PR #151, opened 2023-12-22
Enhance/several minor opts

HYLcool:
- language_id_score_filter supports keeping samples for multiple languages
- np checking to avoid starting too many subprocesses
- modify downloaded and extracted cache dirs that depend on HF_DATASETS_CACHE when it's updated


zhijianma (approved):
LGTM

PR #156, opened 2023-12-26
move data_probe tools to analysis

zhijianma:
move data_probe tools to analysis for importing it easily


drcege (approved):

drcege:
Better to add `seaborn` dependency

PR #157, opened 2023-12-26
* bug fixed: missing attribute dataset_dir when the input dataset path is remote or a mixture of several datasets.

HYLcool:
As the title says.


zhijianma (approved):
LGTM

PR #170, opened 2024-01-04
fix: change datatype of simhash to string, because pyarrow is incompatible with uint64

zhijianma:
As title say.


HYLcool (approved):
LGTM

PR #177, opened 2024-01-15
Feature/audio duration filter

HYLcool:
- Add new OP audio_duration_filter
- Use unified multimodal data loading function with context for multimodal OPs (images & audios)

### DIFF
--- a/configs/config_all.yaml
+++ b/configs/config_all.yaml
@@ -112,8 +112,8 @@ process:
       min_ratio: 0.0                                          # the min ratio of filter range
       max_ratio: 0.9                                          # the max ratio of filter range
   - audio_duration_filter:                                  # Keep data samples whose audios' durations are within a specified range.
-      min_duration: 0                                         # the min audio duration of filter range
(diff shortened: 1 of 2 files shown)


configs/config_all.yaml
```diff
@@ -111,6 +111,10 @@ process:
       tokenization: false                                     # Whether to count the ratio of alphanumeric to the total number of tokens.
       min_ratio: 0.0                                          # the min ratio of filter range
       max_ratio: 0.9                                          # the max ratio of filter range
+  - audio_duration_filter:                                  # Keep data samples whose audios' durations are within a specified range.
+      min_duration: 0                                         # the min audio duration of filter range
+      max_duration: 3600                                      # the max audio duration of filter range
```
drcege, line 116: What is the unit of duration, I guess seconds?  Suggest documenting it more clear (as well as other places).

HYLcool: 👍🏻Yes, it's in seconds. The unit information has been added in the latest commit.

drcege (approved):

PR #191, opened 2024-01-19
Enhance/support prompt for caption generation

HYLcool:
As the title says.

### DIFF
--- a/data_juicer/ops/mapper/generate_caption_mapper.py
+++ b/data_juicer/ops/mapper/generate_caption_mapper.py
@@ -38,6 +38,7 @@ class GenerateCaptionMapper(Mapper):
                  caption_num: PositiveInt = 1,
                  keep_candidate_mode: str = 'random_any',
                  keep_original_sample: bool = True,
+                 prompt: str = None,
                  *args,
                  **kwargs):
         """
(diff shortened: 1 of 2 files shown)


yxdyc (approved):
LGTM

data_juicer/ops/mapper/generate_caption_mapper.py
```diff
@@ -157,9 +164,19 @@ def _process_single_sample(self, ori_sample):
             # generated_text_candidates_single_chunk[i][j] indicates
             # the $i$-th generated candidate for the $j$-th image
 
+            # construct prompts
+            if self.prompt_key \
+                    and isinstance(ori_sample[self.prompt_key], str):
+                # check prompt_key is not None, and it's a str in the sample
+                prompt_texts = [ori_sample[self.prompt_key]] * len(image_chunk)
+            elif self.prompt and isinstance(self.prompt, str):
```
yxdyc, line 172: maybe the line 172 and 168 need to be re-ordered? users may want to quickly change the prompt by overriding the prompt-keys, if both of them are specified. BTW, we can warn users with logger for this case

BeachWang: maybe we need to check if the ori_sample contains this key.

HYLcool: > maybe we need to check if the ori_sample contains this key.

Actually, in line 175, if self.prompt_key doesn't exist in ori_sample, ori_sample[self.prompt_key] will return None, and it's not an instance of str, so the condition will be False.

HYLcool: > maybe the line 172 and 168 need to be re-ordered? users may want to quickly change the prompt by overriding the prompt-keys, if both of them are specified. BTW, we can warn users with logger for this case

A warning will be reported when both `prompt` and `prompt_key` are set in the initialization method.

PR #195, opened 2024-01-23
Docs/bad data exhibition

HYLcool:
As the title says.


HYLcool:
MMC4 dataset is in analysis for these 2 docs.

yxdyc (approved):
The todo item can be finished in next PR: 
[add "bad" data examples from the analysis result of MMC4 dataset for these 2 docs. ]

HYLcool:
- "Bad" data examples from MMC4 have been added to the exhibition
- Fix bugs for mmc4_to_dj/dj_to_mmc4 tools: there might be multiple images matching the same sentence, so we need to support adding multiple image_special_tokens to the same sentence in mmc4_to_dj and parsing them correctly in dj_to_mmc4.

yxdyc (approved):
The modifications lgtm.

PR #202, opened 2024-01-30
Enhance/face area filter

drcege:
* Constrain the detection coordinates within the image so that the filtering ratios can be deterministically controlled between 0 and 1.
* Add debug logging for future inspection when unit test inconsistencies arise.


chenhesen (approved):
LGTM

PR #203, opened 2024-01-30
Enhance/gpu support

drcege:
## GPU Model Loading
- Models are now loaded onto GPUs where possible to maximize efficiency.
  - Currently, this feature is exclusive to Hugging Face models.
  - Adherence to the `CUDA_VISIBLE_DEVICES` setting ensures only designated GPUs are utilized.
  - The `num_proc` parameter is intelligently adjusted for optimal performance with mixed CPU/GPU operations.

## Multiprocessing Strategy
- Will automatically determine the most effective multiprocessing start method.
  - On GPU setups, `forkserver` is the default choice, switching to `spawn` if necessary.
  - For CPU-bound execution, `fork` provides the best compatibility and performance.
- Users retain control over multiprocessing behavior via the `MP_START_METHOD` environment variable, with permissible values including `auto`, `fork`, `forkserver`, and `spawn`.

## New Operator Development
- When developing GPU-accelerated operators:
  - Declare GPU usage with `self._accelerator = 'cuda'`.
  - Both `compute_stats` and `process` methods should now include an optional `rank` parameter (`rank=None`) to support multi-GPU setups.


data_juicer/__init__.py
```diff
@@ -1 +1,58 @@
...
+
+
+def cuda_device_count():
+    return __CUDA_COUNT
+
+
+def setup_mp():
+    method = os.getenv('MP_START_METHOD', 'auto').lower()
+    if method == 'auto':
+        import torch
```
HYLcool, line 23: If so, dependency `torch` might need to be moved to minimal_requires?

drcege: I have replaced the torch method with a custom function, how about that?

drcege:
Added GPU development guide.

HYLcool (approved):
LGTM 🎉

PR #200, opened 2024-01-29
image diffusion mapper

BeachWang:
generate and augment the images by stable diffusion model.

PR #217, opened 2024-02-22
fix single gpu case

drcege:
There is a special case where `datasets.map` does not start multiprocessing when set to run on a single GPU.

Previously, `get_model` would only use the GPU during multiprocessing, leading to the model not being transferred to CUDA in the single-process setting.


HYLcool (approved):
LGTM.

Conflicts in this file between #200 and this PR should be resolved with @BeachWang

Issue #241, opened 2024-03-11
[Bug]: video split by duration mapper return non-exist video


garyzhang99:
### Before Reporting 报告之前

- [X] I have pulled the latest code of main branch to run again and the bug still existed. 我已经拉取了主分支上最新的代码，重新运行之后，问题仍不能解决。

- [X] I have read the [README](https://github.com/alibaba/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process. (Otherwise, we recommend that you can ask a question using the Question template) 我已经仔细阅读了 [README](https://github.com/alibaba/data-juicer/blob/main/README_ZH.md) 上的操作指引，并且在安装过程中没有错误发生。（否则，我们建议您使用Question模板向我们进行提问）

### Search before reporting 先搜索，再报告

- [X] I have searched the Data-Juicer [issues](https://github.com/alibaba/data-juicer/issues) and found no similar bugs. 我已经在 [issue列表](https://github.com/alibaba/data-juicer/issues) 中搜索但是没有发现类似的bug报告。

### OS 系统

Linux

### Installation Method 安装方式

from source

### Data-Juicer Version Data-Juicer版本

latest v0.2.0

### Python Version Python版本

3.8

### Describe the bug 描述这个bug

In VideoSplitByDurationMapper, when the video duration is 10.01 second, and the spilt_duration parameter is 2 second, the returned split_video_keys contain 6 videos, but there are only 5 videos generated.

### To Reproduce 如何复现

1. Modify Demo config:
process:
  - video_split_by_duration_mapper:       
      split_duration: 2                                   
      min_last_split_duration: 0.0                  
      keep_original_sample: true   
  - video_resize_aspect_ratio_mapper:
      min_ratio: 1
      max_ratio: 1.1
      strategy: increase
2. Process a video that is 10 second long.

PR #243, opened 2024-03-11
fix cut video bugs (#241)

chenhesen:
An excessively small time window may result in video cut failure
add appropriate checks and warnings in the code.


HYLcool (approved):
LGTM 👍🏻

PR #244, opened 2024-03-11
Demos/op insight update

zhijianma:
1. add deduplicator 
2. fix diversity demo
3. add clear cache
4. support batched op


HYLcool (approved):
LGTM

PR #245, opened 2024-03-12
Update sphinx docs for v0.2.0

pan-x-c:
As the title says.


yxdyc (approved):
LGTM. BTW, should we also modify config of pre-commit accordingly? It seems like some style in-consistency between linter and sphinx?

pan-x-c:
> LGTM. BTW, should we also modify config of pre-commit accordingly? It seems like some style in-consistency between linter and sphinx?

No need for now.

PR #252, opened 2024-03-14
Dev/nsfw filter

BeachWang:
Add to filter ops: filter samples by NSFW scores of images and videos.


tests/ops/filter/test_image_nsfw_filter.py
```diff
@@ -0,0 +1,100 @@
...
+        tgt_list = [{
+            'images': [self.img2_path, self.img3_path]
+        }]
+
+        dataset = Dataset.from_list(ds_list)
+        op = ImageNSFWFilter(hf_nsfw_model=self.hf_nsfw_model,
+                            score_threshold=0.0005,
+                            any_or_all='all')
+        self._run_filter(dataset, tgt_list, op)   
+
```
HYLcool, line 98: Do we need to test the op in the parallel mode (np > 1)?

tests/ops/filter/test_video_nsfw_filter.py
```diff
@@ -0,0 +1,172 @@
...
+            'videos': [self.video2_path, self.video3_path]
+        }]
+
+        dataset = Dataset.from_list(ds_list)
+        op = VideoNSFWFilter(hf_nsfw_model=self.hf_nsfw_model,
+                            score_threshold=0.1,
+                            frame_sampling_method='all_keyframes',
+                            any_or_all='all')
+        self._run_filter(dataset, tgt_list, op)   
+
```
HYLcool, line 170: Do we need to test the op in the parallel mode (np > 1)?

tests/ops/filter/test_image_nsfw_filter.py
BeachWang: test in multi process passed

tests/ops/filter/test_video_nsfw_filter.py
BeachWang: test in multi process passed

HYLcool (approved):
Others LGTM. Some conflicts might need to be resolved.

PR #253, opened 2024-03-14
Op/video face blur mapper

drcege:
* add `video_face_blur_mapper`: currently detect faces in each frame, which has drawbacks in efficiency and tracking (performance optimization might be considered in the future)
* improve `process_each_frame`: will not save a copy if videos are not modified, and return the right filepath


data_juicer/utils/mm_utils.py
```diff
@@ -364,6 +367,8 @@ def process_each_frame(input_video: Union[str, av.container.InputContainer],
         for packet in container.demux(input_video_stream):
             for frame in packet.decode():
                 new_frame = frame_func(frame)
+                if new_frame != frame:
```
HYLcool, line 370: Will it be better to compare frames only on their image contents? Frames have many other attributes ([ref](https://pyav.org/docs/stable/api/video.html#module-av.video.frame)), some of them are unrelated to visual contents.

drcege: This is merely for a quick check (`__eq__` should default to `is`), as byte-level content comparison may incur significant overhead. The `frame_func` is expected to directly return a reference to the input frame when no modification is needed.

HYLcool: gotcha~

HYLcool (approved):
Please resolve conflicts from the main branch before merging. Others LGTM.

PR #295, opened 2024-04-15
Fix: avoid extracting empty frame list

HYLcool:
* Fix: avoid extracting empty key frame list
* Reopen Unittest for light OPs. Heavy OPs should be tested locally before pushed.


yxdyc (approved):
LGTM

PR #304, opened 2024-04-24
Add more unittest 

pan-x-c:
1. setup local actions runner on a GPU machine
2. use docker-compose to setup a cluster
3. add more unit tests
   - single process
   - multi process
   - multi process with GPU
   - ray with CPU
   - ray with GPU


.github/workflows/docker/docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    deploy:
+      resources:
+        reservations:
+          devices:
+          - driver: nvidia
+            device_ids: ['0', '1']
+            capabilities: [gpu]
+
+  ray-worker:
+    image: data-juicer-unittest:0.2.0
```
yxdyc, line 33: In the future, we may change it into `data-juicer-unittest:latest` and maintain a latest image

data_juicer/utils/unittest_utils.py
```diff
@@ -2,11 +2,30 @@
...
+from datasets import Dataset
+
+from data_juicer.ops import Filter
+from data_juicer.utils.constant import Fields
 from data_juicer.utils.registry import Registry
 
 SKIPPED_TESTS = Registry('SkippedTests')
 
 
+def TEST_TAG(*tags):
```
yxdyc, line 17: There are two type of tags, while the PR's description mentioned 5 types of test scenarios. So what is the mapping relationship? 
```
single process
multi process
multi process with GPU
ray with CPU
ray with GPU
```

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,60 @@
...
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  unittest-single:
+    runs-on: [self-hosted, linux]
```
yxdyc, line 18: By now, does our self-hosted machine satisfy the requirements from security colleagues?

data_juicer/utils/unittest_utils.py
pan-x-c: For now, there is no distinction between CPU and GPU testing, and the corresponding configuration will be added in the future when needed.

.github/workflows/unittest.yml
pan-x-c: Nope

drcege (approved):

PR #306, opened 2024-04-25
Minor modifications

HYLcool:
* download imagededup models when they are used instead of imported
* set the minimum version limitation for torch


setup.py
```diff
@@ -41,6 +41,8 @@ def get_install_requirements(require_f_paths, env_dir='environments'):
     'tools':
     get_install_requirements(
         ['preprocess_requires.txt', 'quality_classifier_requires.txt']),
+    'sandbox':
```
yxdyc, line 44: It is easier to cause more conflict if we put all utilized packages into one env? I recommend to a use-case based split, such as `sandbox_modelscope`, `sandbox_llava`, `sandbox_llamafactory`, `sandbox_easyanimate`  

If this suggestion adopted, we may need also add utility similar to `with AvailabilityChecking()` for each Sandbox's component

yxdyc (approved):

PR #311, opened 2024-05-08
bug fixed: params from base class can't be parsed in some OPs

HYLcool:
By the way, a new test case is added to test param parsing


yxdyc (approved):
LGTM. 

Explanation: The modified parts originally cause assertion error for an constant node in AST building, we fix this problem in this PR

PR #317, opened 2024-05-20
Add source tag

chenhesen:
Add source tags to the newly generated images, videos, and audio in some mapper op.


yxdyc (approved):
The unit test needs to be fixed. Others LGTM.

PR #318, opened 2024-05-23
Dev/fix video filter for bench

BeachWang:
1. fix video filter OPs for bench experiments.
2. fix words_num_filter: use both word_num_filter and words_num_filter in the old version.


HYLcool (approved):
LGTM

PR #319, opened 2024-05-23
support-mul-pattern

seanzhang-zhichen:
replace_content_mapper支持传递多个正则表达式

### DIFF
--- a/data_juicer/ops/mapper/replace_content_mapper.py
+++ b/data_juicer/ops/mapper/replace_content_mapper.py
@@ -1,5 +1,6 @@
+from typing import List, Union
+
 import regex as re
-from typing import Union, List
 
 from ..base_op import OPERATORS, Mapper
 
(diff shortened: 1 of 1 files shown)


data_juicer/ops/mapper/replace_content_mapper.py
```diff
@@ -9,30 +11,48 @@ class ReplaceContentMapper(Mapper):
...
 
-        sample[self.text_key] = re.sub(pattern=self.pattern,
-                                       repl=self.repl,
-                                       string=sample[self.text_key],
-                                       flags=re.DOTALL)
+        for i, pattern in enumerate(self.compiled_patterns):
+            if isinstance(self.repl, list) and i < len(self.repl):
+                replacement = self.repl[i]
+            else:
+                replacement = self.repl
```
HYLcool, line 53: What if `self.repl` is a list but `i >= len(self.repl)`?

HYLcool (approved):
LGTM. Thanks for you contribution!👍🏻

PR #320, opened 2024-05-24
Switch face detection to OpenCV

drcege:
Switch face detections to `opencv` implementation. Remove dependency on `dlib`.

**Note:** All opencv-related OPs cannot run in multiprocessing with `fork` mode, which should be considered during refactoring.


HYLcool (approved):
LGTM

PR #325, opened 2024-06-05
Sample level error catching

garyzhang99:
The PR here is an example of rewriting sample error catching.

The current data-juicer abruptly stops the program when an error arises due to an issue with a specific sample, which requires a direct rerun, and no outputs are exported. As shown in the image:
image

After modification, errors caused by a specific sample will not interrupt the program; instead, exception handling will take place. The subsequent steps will continue to run and export the final outputs. As shown in the image:
image

FUTURE TODOs:

- [ ] Expand to filters
- [ ] Optimize batch efficiency
- [ ] Improve the transmission of error messages

### DIFF
--- a/data_juicer/ops/base_op.py
+++ b/data_juicer/ops/base_op.py
@@ -2,11 +2,48 @@ import copy
 
 import pandas as pd
 import pyarrow as pa
+from loguru import logger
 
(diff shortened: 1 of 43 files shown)


drcege:
Closed as impelemented in #336

PR #334, opened 2024-06-25
enhance video_motion_score_filter

drcege:
- Added `target_size` parameter to allow downsampling of frames before computing optical flow. 
    - This can significantly speed up computation.
    - Note that this results in different motion scores, but maintains the order of magnitude.
- Added `normalize` parameter to control whether to normalize motion scores to the `[0,1]` range. 
    - This might be useful when comparing videos with significantly different resolutions. 
    - Note that normalized values can be very small (e.g., on the order of 1e-3), so please set thresholds carefully.
- Enhanced exception handling to improve the robustness of OpenCV video operations.
- Increased the maximum line length for `pre-commit` to a more modern 119 characters.


ycwfs (approved):

data_juicer/ops/filter/video_motion_score_filter.py
```diff
@@ -128,18 +161,18 @@ def compute_stats(self, sample, context=False):
...
-                unique_motion_scores[video_key] = -1
-            else:
-                unique_motion_scores[video_key] = np.mean(video_motion_scores)
+            # set to -1 upon loop break
+            unique_motion_scores[video_key] = np.mean(video_motion_scores or [-1])
 
         sample[Fields.stats][StatsKeys.video_motion_score] = [
             unique_motion_scores[key] for key in loaded_video_keys
         ]
+        print(sample)
```
yxdyc, line 175: we can remove this print after debugging

yxdyc (commented):
LGTM, we can merge it after fixing conflict and minor comments

PR #336, opened 2024-06-28
Refactor OP & Dataset

drcege:
- [x] Decouple operator logic from executor. Enabled new interfaces: 
    - [x] `dataset = dataset.process(op)`
    - [x] `dataset = dataset.process([op1, op2])`
    - [x] `dataset = op(dataset)`
    - [x] `dataset = op.run(dataset)`
- [ ] Integrate exception handling
  - [ ] #331 
  ⚠️ Be careful with non-sharable `loguru.logger`
  - [x] #325 
- [x] Work with Ray.
  - [ ] Some warnings to be checked
  - [x] [skipped] Ray `dataset.filter` does not have batched version
- [x] Work with batched op. All fault-tolerant `dataset.map` functions must be batched, the following methods are now wrapped:
  - [x] `Mapper.process`
  - [x] `Filter.compute_stats`
  - [x] `Deduplicator.compute_hash`


data_juicer/ops/base_op.py
```diff
@@ -49,6 +53,13 @@ def __init__(self, *args, **kwargs):
     def process(self, *args, **kwargs):
         raise NotImplementedError
 
+    def runtime_np(self):
+        return calculate_np(self._name, self.mem_required, self.cpu_required,
+                            self.num_proc, self.use_cuda)
```
garyzhang99, line 58: typo here, should be self.use_cuda()

yxdyc (approved):
Nice work, LGTM

PR #343, opened 2024-07-05
fix vid_cap_from_frm_args None error

simplaj:
when use "video_captioning_from_summarizer_mapper"  and set  "vid_cap_from_frm_args"  to "null", will produce bug.


drcege (approved):
LGTM

Issue #345, opened 2024-07-07
Unable to process data with ray executor on Kubernetes Ray Cluster:


Fatima-0SA:
### Before Reporting 报告之前

- [X] I have pulled the latest code of main branch to run again and the bug still existed. 我已经拉取了主分支上最新的代码，重新运行之后，问题仍不能解决。

- [X] I have read the [README](https://github.com/alibaba/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process. (Otherwise, we recommend that you can ask a question using the Question template) 我已经仔细阅读了 [README](https://github.com/alibaba/data-juicer/blob/main/README_ZH.md) 上的操作指引，并且在安装过程中没有错误发生。（否则，我们建议您使用Question模板向我们进行提问）

### Search before reporting 先搜索，再报告

- [X] I have searched the Data-Juicer [issues](https://github.com/alibaba/data-juicer/issues) and found no similar bugs. 我已经在 [issue列表](https://github.com/alibaba/data-juicer/issues) 中搜索但是没有发现类似的bug报告。

### OS 系统

Ubuntu

### Installation Method 安装方式

from source, for distributed processing

### Data-Juicer Version Data-Juicer版本

v0.2.0

### Python Version Python版本

3.10.14

### Describe the bug 描述这个bug

I have created a RayCluster deployed on Kubernetes within a VM -ray version 2.31.0-. I have forwarded the head ports:
- 10001 (for Ray Client connection) and,
-  8265 (for the dashboard view). 

I am trying to run the process_on_ray demo from a local docker container that has data juicer installed on it.  For processing on ray, I have only changed the ray address from `'auto'` to: `ray_address: 'ray://host.docker.internal:10001'` and reduced the processing steps in the demo to include only `average_line_length_filter`. 

Running the above, I managed to get the connection initialization successfully, and I can see the jobs are scheduled within the ray dashboard, however, when it comes to reading the JSON file as a dataset, and trying to get the columns of the dataset, I face RuntimeError: Global Node is not initialized. This applies to any kind of processing step I want to perform on the dataset object. 

Any insight about the global node initialization error? or is there any specific config. to be done when deploying the RayCluster?  

### To Reproduce 如何复现

1. Deploy RayCluster on kubernetes
2. Forward the head pod to port 10001
3. Create a Docker container that has data-juicer
4. Change process_on_ray configuration to be similar to the config mentioned below
[...]

PR #348, opened 2024-07-11
Refactor dataset

pan-x-c:
- Add `DJDataset` as the base interface of all datasets in data-juicer.
- Add `RayDataset` as a sub-class of `DJDataset`
- Simplify `RayExecutor`
- update version of RAY to 2.31.0
- fix the ray-mode dataset loading method according to #345


data_juicer/core/ray_data.py
```diff
@@ -0,0 +1,123 @@
...
+    if not use_cuda() or not op._accelerator == 'cuda':
+        return 0
+    proc_per_gpu = op_proc / cuda_device_count()
+    return 1.0 / proc_per_gpu
+
+
+class RayDataset(DJDataset):
+
+    def __init__(self, dataset: Dataset, dataset_path: str, cfg) -> None:
+        dataset = preprocess_dataset(dataset, dataset_path, cfg)
```
drcege, line 81: `dataset` should be `self.data`?

PR #349, opened 2024-07-11
fix ray batch-sample format

drcege:
Temporary fix for the format issues caused by implicit conversion of lists to numpy arrays in ray batch mode.


drcege:
Included in #359

PR #351, opened 2024-07-11
Fix several bugs for image OPs

HYLcool:
- Rename "id" with the correct var name "sid" in `dj_to_llava.py` tool.
- Update the GPU Mem required by image captioning and diffusion mappers.
- Allow loading truncated images for some too-large images.
- For multimodal deduplicators, when considering hashes of texts, compute their hashes first to avoid empty text hashes. Add corresponding unit tests as well.
- Downgrade the log level of the model moving on GPUs from "info" to "debug".


drcege (approved):
LGTM

drcege:
@HYLcool  Could you take a look at the following warning when you have a moment?
```bash
Parameter 'function'=<function DocumentMinhashDeduplicator.compute_hash at 0x7fa3417db8b0> of the transform datasets.arrow_dataset.Dataset._map_single couldn't be hashed properly, a random hash was used instead. Make sure your transforms and parameters are serializable with pickle or dill for the dataset fingerprinting and caching to work. If you reuse this transform, the caching mechanism will consider it to be different from the previous calls and recompute everything. 
```

PR #359, opened 2024-07-17
Refactor OP & Dataset (from #336)

yxdyc:
- [x] Decouple operator logic from executor. Enabled new interfaces: 
    - [x] `dataset = dataset.process(op)`
    - [x] `dataset = dataset.process([op1, op2])`
    - [x] `dataset = op(dataset)`
    - [x] `dataset = op.run(dataset)`
  ⚠️ Be careful with non-sharable `loguru.logger`
  - [x] #325 
- [x] Work with Ray.
  - [x] [skipped] Ray `dataset.filter` does not have batched version
- [x] Work with batched op. All fault-tolerant `dataset.map` functions must be batched, the following methods are now wrapped:
  - [x] `Mapper.process`
  - [x] `Filter.compute_stats`
  - [x] `Deduplicator.compute_hash`

Next Steps on following PRs:
  - Check warnings in Ray mode
  - improve exception handling #331 
  - Docs reflecting the new nice features from this refactor, interfaces, fault-tolerance, ...

PR #365, opened 2024-07-24
correct num_proc logic

drcege:
Currently, when OP-specific `num_proc` argument is not set, it will not automatically fall back to global `np`.

- This is a bug accidentally introduced in a previous commit, now fixed
- Also, fix an edge case where the `rank` is None in the OCR operator when `np=1`


garyzhang99:
In the previous refactor PR https://github.com/modelscope/data-juicer/pull/359, we set all op's ray mode to **task**, which will cause CUDA OOM error. 
We need to add back ray **actor** mode for op that use gpu.

garyzhang99 (approved):

PR #374, opened 2024-07-29
fix video memory leak

BeachWang:
as the title say

### DIFF
--- a/data_juicer/utils/mm_utils.py
+++ b/data_juicer/utils/mm_utils.py
@@ -319,8 +319,8 @@ def cut_video_by_seconds(
 
     # close the output videos
     if isinstance(input_video, str):
-        container.close()
-    output_container.close()
+        close_video(container)
+    close_video(output_container)
     if not os.path.exists(output_video):
         logger.warning(f'This video could not be successfully cut in '
                        f'[{start_seconds}, {end_seconds}] seconds. '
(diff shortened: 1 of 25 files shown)


data_juicer/utils/mm_utils.py
```diff
@@ -786,3 +786,13 @@ def parse_string_to_roi(roi_string, roi_type='pixel'):
...
     return None
+
+
+def close_video(container):
+    """
+    Close the video stream and container to avoid memory leak.
+
+    :param container: the video container.
+    """
+    container.streams.video[0].close(strict=False)
```
garyzhang99, line 797: Not an issue, but I wander that will there be a video[1]?

tools/mm_eval/inception_metrics/dataset.py
```diff
@@ -117,6 +118,7 @@ def __post_init__(self):
                     num_samples_from_source = total_frame_num - self.seq_length + 1
                     for start_frame in range(0, num_samples_from_source):
                         self.start_frames.append((video_path, start_frame, num_samples_from_source))
+                    container.streams.video[0].close()
```
garyzhang99, line 121: Why not use `close_video(container)` here, is it because you don't need(strict=False) or because of other reasons?

BeachWang: because not use load_video here, and strict=False is required here

data_juicer/utils/mm_utils.py
BeachWang: a loop has been added here

tools/mm_eval/inception_metrics/dataset.py
BeachWang: have changed to use `load_video` and `close_video`

yxdyc (approved):

PR #389, opened 2024-08-19
+ add use_cuda for get_model funcs in two OPs

HYLcool:
As the title says.

This PR solve the problem that some OP can not make use of CUDA to speed up the loaded models.

### DIFF
--- a/data_juicer/ops/filter/video_aesthetics_filter.py
+++ b/data_juicer/ops/filter/video_aesthetics_filter.py
@@ -159,7 +159,9 @@ class VideoAestheticsFilter(Filter):
 
             if len(frame_images) > 0:
                 # compute aesthetics_scores
-                model, processor = get_model(self.model_key, rank=rank)
+                model, processor = get_model(self.model_key,
+                                             rank=rank,
+                                             use_cuda=self.use_cuda())
                 inputs = processor(images=frame_images,
                                    return_tensors='pt').to(model.device)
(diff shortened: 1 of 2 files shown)


yxdyc (approved):
This PR is just a quick fix. Later we may add auto inspect for all model-based ops whose `_accelerator='cuda'`, with similar codes below:
```data-juicer/data_juicer/utils/model_utils.py

import inspect
......

def get_model(model_key=None, rank=None):
    if model_key is None:
        return None

    global MODEL_ZOO
    if model_key not in MODEL_ZOO:
        logger.debug(
            f'{model_key} not found in MODEL_ZOO ({mp.current_process().name})'
        )
        MODEL_ZOO[model_key] = model_key()
    
    frame = inspect.currentframe().f_back
    caller_locals = frame.f_locals
    caller_class = None
    for var_name, var_value in caller_locals.items():
        if isinstance(var_value, type):
            caller_class = var_value
            break
        if hasattr(var_value, '__class__'):
            caller_class = var_value.__class__
            break

    if caller_class is not None and caller_class.__name__ == 'DJ_operator': # to be corrected
        use_cuda = True               
    else: 
        use_cuda = False
    if use_cuda:
        rank = 0 if rank is None else rank
        rank = rank % cuda_device_count()
        move_to_cuda(MODEL_ZOO[model_key], rank)
    return MODEL_ZOO[model_key]

```

PR #391, opened 2024-08-20
Fix an edge case when the current configuration has fewer OPs than the checkpoint

### LINKED ISSUE #337: potential bug of checkpointing
As the title suggests, the issue lies in the following code. 

https://github.com/modelscope/data-juicer/blob/c85e024c718b8e2dab967de478d15dc5db3dd907/data_juicer/utils/ckpt_utils.py#L86-L101

When the new `process_list` is shorter than `op_record`, Python does not raise an error for out-of-range indices but rather truncates to the maximum available length, thus `len(prefix_process) < len(self.op_record)`. Similarly, the `zip` function terminates at the shorter iterable's length. This results in the `check_ops_to_skip` function incorrectly assuming that the recorded operators match the prefix of the current operators list. 

 Is that the case? @HYLcool @yxdyc

drcege:
close #337

### DIFF
--- a/data_juicer/utils/ckpt_utils.py
+++ b/data_juicer/utils/ckpt_utils.py
@@ -87,10 +87,9 @@ class CheckpointManager:
         process_op_num = len(self.process_list)
         if process_op_num < recorded_op_num:
             logger.warning(
-                f'Current config ops ({process_op_num}) are fewer '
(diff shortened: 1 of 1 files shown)


HYLcool (approved):
LGTM 👍🏻

PR #393, opened 2024-08-21
Add image_pair_similarity_filter

Qirui-jiao:
Calculate the cosine similarity of CLIP image features for paired images, and filter the samples based on this value. 
Hyperparameters: 
- min_score – the min similarity score of filter range.
- max_score – the max similarity score of filter range.


tests/ops/filter/test_image_pair_similarity_filter.py
```diff
@@ -0,0 +1,60 @@
...
+            'images': ["./crayon.jpg", "./ipod.jpg"]
+        }]
+
+
+        dataset = Dataset.from_list(ds_list)
+        op = ImagePairSimilarityFilter(hf_clip="clip-vit-base-patch32",
+                                       any_or_all='any',
+                                       min_score=0.85,
+                                       max_score=0.98)
+        self._run_filter(dataset, op)
```
HYLcool, line 56: Please use a target dataset to check the filter results are expected like unit tests for other Filters. Besides, you can add the test images to `tests/ops/data`, otherwise the unit tests on the server will not be passed because of "image not found" error

Refer: 
1. image paths: https://github.com/modelscope/data-juicer/blob/22834baea64a4def7348ecc18b1cde4e68ec738d/tests/ops/filter/test_image_text_similarity_filter.py#L15-L19
2. target checking: https://github.com/modelscope/data-juicer/blob/22834baea64a4def7348ecc18b1cde4e68ec738d/tests/ops/filter/test_image_text_similarity_filter.py#L41

tests/ops/filter/test_image_pair_similarity_filter.py
```diff
@@ -0,0 +1,60 @@
...
+from data_juicer.core.data import NestedDataset as Dataset
+
+from data_juicer.ops.filter.image_pair_similarity_filter import ImagePairSimilarityFilter
+from data_juicer.utils.constant import Fields
+from data_juicer.utils.mm_utils import SpecialTokens
+from data_juicer.utils.unittest_utils import (SKIPPED_TESTS,
+                                              DataJuicerTestCaseBase)
+
+# These tests have been tested locally.
+@SKIPPED_TESTS.register_module()
```
HYLcool, line 13: This decorator can be removed now.

data_juicer/ops/filter/image_pair_similarity_filter.py
```diff
@@ -0,0 +1,114 @@
...
+
+        image_list = []
+        for temp_key in images.keys():
+            image_list.append(images[temp_key])
+        image_tensors = processor.image_processor(
+            image_list, return_tensors='pt')['pixel_values']
+        image1_batch_feature = model.get_image_features(
+            image_tensors[0].unsqueeze(0).to(model.device))
+        image2_batch_feature = model.get_image_features(
+            image_tensors[1].unsqueeze(0).to(model.device))
```
HYLcool, line 91: The number of images in the image_list needs to be checked. For example, if there is only 1 image in the list, we can throw a ValueError for this sample.

tests/ops/filter/test_image_pair_similarity_filter.py
HYLcool: Same issue for all your contributed OPs.

tests/ops/filter/test_image_pair_similarity_filter.py
HYLcool: Same issue for all your contributed OPs.

HYLcool (approved):
LGTM

PR #394, opened 2024-08-21
[Ready] Add image_segment_mapper

Qirui-jiao:
Perform segment-anything on images (with FastSAM) and return the bounding box values.
Hyperparameters:
- imgsz: image resolution after image resizing
- conf: confidence score threshold
- iou: IoU (Intersection over Union) score threshold


data_juicer/utils/model_utils.py
```diff
@@ -553,6 +553,12 @@ def prepare_recognizeAnything_model(
     return model
 
 
+def prepare_fastsam_model(pretrained_model_name_or_path):
+    from ultralytics import FastSAM
+
+    return FastSAM(pretrained_model_name_or_path)
```
HYLcool, line 559: how about downloading the checkpoint automatically for users? If not, we'd better add a comment in config_all.yaml and doc string of this OP to tell users how to download them manually.

data_juicer/ops/mapper/segment_mapper.py
```diff
@@ -0,0 +1,88 @@
...
+        images = {}
+        for loaded_image_key in loaded_image_keys:
+            if loaded_image_key not in images:
+                # avoid loading the same images
+                image = load_image(loaded_image_key)
+                images[loaded_image_key] = image
+
+        model = get_model(self.model_key, rank=rank, use_cuda=self.use_cuda())
+
+        generated_samples['bboxes'] = []
```
HYLcool, line 73: For these new first-level fields, we can add a new field name for them in the Fields class of `utils/constant.py`.

Refer:
1. First-level fields for tags by two mappers: https://github.com/modelscope/data-juicer/blob/22834baea64a4def7348ecc18b1cde4e68ec738d/data_juicer/utils/constant.py#L19-L21
2. How these two tag fields are used in the mapper: https://github.com/modelscope/data-juicer/blob/22834baea64a4def7348ecc18b1cde4e68ec738d/data_juicer/ops/mapper/video_tagging_from_frames_mapper.py#L118

data_juicer/ops/mapper/segment_mapper.py
```diff
@@ -0,0 +1,88 @@
...
+from data_juicer.utils.mm_utils import load_image
+from data_juicer.utils.model_utils import get_model, prepare_model
+
+OP_NAME = 'segment_mapper'
+
+with AvailabilityChecking(['torch', 'transformers', 'simhash-pybind'],
+                          OP_NAME):
+    import simhash  # noqa: F401
+    import torch
+    import transformers  # noqa: F401
```
HYLcool, line 15: `simhash` and `transformers` are not used in this OP, so we can remove them from imports here and AvailabilityChecking list above.

On the other hand, `ultralytics` is used implicitly in this OP when loading the model. So we need add AvailabilityChecking and import for this library here.

tests/ops/mapper/test_segment_mapper.py
```diff
@@ -0,0 +1,46 @@
...
+    text_key = 'text'
+
+    def _run_segment_mapper(self, enable_vllm=False):
+        op = SegmentMapper(
+            fastsam_path='FastSAM-x.pt',
+        )
+
+        img1_path = './crayon.jpg'
+        img2_path = './ipod.jpg'
+        img3_path = './0_19_0_0.jpg'
```
HYLcool, line 26: resolve similar issues mentioned in #393

data_juicer/utils/model_utils.py
Qirui-jiao: I set the default model path to 'FastSAM-x.pt,' and when the user does not provide a model path, **the ultralytics library will automatically downloads the model**. When the model weight are not found in the specified path, the ultralytics library will also automatically download it.

drcege:
Update: Merged `main` and adapted to the latest development paradigm.

HYLcool:

PR #396, opened 2024-08-22
Enhance/ckpt

drcege:
Enhance the checkpointing logic so that a checkpoint check is performed regardless of whether the OPs succeed or fail, occurring before the export.


HYLcool (approved):
LGTM

PR #397, opened 2024-08-22
update spacy to deal conflict with ms-swift

BeachWang:
Have test related OP, the spacy can be the latest version.

### DIFF
--- a/.github/workflows/docker/docker-compose.yml
+++ b/.github/workflows/docker/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   ray-head:
-    image: data-juicer-unittest:0.2.1
+    image: data-juicer-unittest:0.2.2
     pull_policy: never
     command: ray start --head --dashboard-host 0.0.0.0 --include-dashboard true --block
     environment:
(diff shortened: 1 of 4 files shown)


drcege:
Is the model version being handled correctly? (Ref: https://spacy.io/models)

https://github.com/modelscope/data-juicer/blob/213f7f8aef078395eaefd89817b098eeb94a45b4/data_juicer/utils/model_utils.py#L409

BeachWang:
> Is the model version being handled correctly? (Ref: https://spacy.io/models)
> 
> https://github.com/modelscope/data-juicer/blob/213f7f8aef078395eaefd89817b098eeb94a45b4/data_juicer/utils/model_utils.py#L409

The higher version of `spacy` is compatible with lower version models, and there is no `zh_core_web_md` model, so the model will temporarily remain at `3.5.0`.

drcege:
Please check the following link: [compatibility.json](https://github.com/explosion/spacy-models/blob/master/compatibility.json). `sm`, `md`, and `lg` are simply indicators of model sizes.

@HYLcool, could you also take a look at how we can adaptively handle versioning?

drcege:
Pin the version to 3.7.0 and support extracting native `tar.gz` format.

Ideally, we should be able to automatically download the required packages via `spacy.cli.download`. This could potentially be considered in conjunction with issue #398.

drcege (approved):

PR #401, opened 2024-08-22
Add sentence_augmentation_mapper

Qirui-jiao:
Augment sentences using LLMs.
Hyperparameters:
- system_prompt: system prompt;
- task_sentence: the instruction for the current task;
- max_new_tokens: the maximum number of new tokens generated by the model;
- sampling_params: sampling parameters for text generation.


data_juicer/ops/mapper/sentence_augmentation_mapper.py
```diff
@@ -0,0 +1,138 @@
...
+    def process(self, sample=None, rank=None):
+
+        if 'vicuna' in self.hf_model:
+            if torch.cuda.is_available():
+                model, processor = get_model(self.model_key,
+                                             rank=rank,
+                                             use_cuda=True)
+            else:
+                model, processor = get_model(self.model_key, rank=rank)
+
```
zhijianma, line 74: Perhaps you could optimize the code as follows
```py
model, processor = get_model(model_key=self.model_key,
                                                    rank=rank,
                                                    use_cuda=self.use_cuda())
```

data_juicer/ops/mapper/sentence_augmentation_mapper.py
```diff
@@ -0,0 +1,138 @@
...
+        self.model_key = prepare_model(model_type='huggingface',
+                                       pretrained_model_name_or_path=hf_model)
+        self.temperature = temperature
+        self.top_p = top_p
+        self.num_beams = num_beams
+        self.task_sentence = task_sentence
+
+    def process(self, sample=None, rank=None):
+
+        if 'vicuna' in self.hf_model:
```
zhijianma, line 67: The only difference between different `hf_models` is the template for the input prompt.
 Could we create `input_prompt ` in branch while keeping the rest of the code shared?

data_juicer/ops/mapper/sentence_augmentation_mapper.py
```diff
@@ -0,0 +1,138 @@
...
+            if torch.cuda.is_available():
+                model, processor = get_model(self.model_key,
+                                             rank=rank,
+                                             use_cuda=True)
+            else:
+                model, processor = get_model(self.model_key, rank=rank)
+
+            input_prompt = self.system_prompt + " USER: Here \
+                is a sentence: \"" + sample[
+                self.text_key] + "\". " + self.task_sentence
```
zhijianma, line 77: Please make sure:
- `task_sentence` takes `None` as default value,  so it may lead to a string + None error when it is not configured.
- `sample[self.text_key]` will utilize all the text content, specifically at the document level. It would be helpful to add a description of the usage scenarios.

zhijianma (approved):

data_juicer/ops/mapper/sentence_augmentation_mapper.py
```diff
@@ -0,0 +1,138 @@
...
+            response = model.generate(**inputs,
+                                      max_new_tokens=self.max_new_tokens,
+                                      eos_token_id=processor.eos_token_id,
+                                      top_p=self.top_p,
+                                      temperature=self.temperature,
+                                      num_beams=self.num_beams)
+            input_token_len = inputs.input_ids.shape[1]
+            n_diff_input_output = (inputs.input_ids !=
+                                   response[:, :input_token_len]).sum().item()
+            if n_diff_input_output > 0:
```
zhijianma, line 129: This needs to be handled carefully, as the size of the vocabulary may lead to different tokenizations, thereby affecting the length of `ids`. Would it be better to make the judgment after decoding?

zhijianma (approved):

data_juicer/ops/mapper/sentence_augmentation_mapper.py
Qirui-jiao: Sorry,  I don't quite understand.  I use the same tokenizer for both encoding and decoding, why does the size of the vocabulary differ? Based on my experiments, making the judgment before decoding has proven effective for models like Vicuna and Qwen2. Additionally, I've noticed that repositories like LLaVA also employ this method.

If the judgment is made after decoding, special tokens would be decoded as well, which means I wouldn't be able to use the original input directly for making the judgment.

zhijianma: Here is just a hypothetical case:
original text:    a book is near a cat and a dog.
output text:  a book is near a kangaroo and a dog.
and `kangaroo` maybe have more idx than `cat`.

Qirui-jiao: But the entire response is: 

A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions. User: Here is a sentence: "a book is near a cat and a dog." Please replace one entity in this sentence with another entity, such as an animal, a vehicle, or a piece of furniture. Please only answer with the replaced sentence. ASSISTANT: **a book is near a kangaroo and a dog.** (The bolded sentence indicates newly generated content)

**The decoding for inputs_ids only covers the sentences from "A chat" to "ASSISTANT:".**

zhijianma: As you mentioned, is the prefix from 'A chat' to 'ASSISTANT:' in the response the same as the input prompt? If so, can it be cut off directly and then decoded?

HYLcool:

PR #402, opened 2024-08-23
Add turbo mode 

drcege:
Add `turbo` mode to allow disabling fault tolerance and maximize processing speed; future optimizations in batch processing should aim to address the performance gap until the option becomes impactless.


yxdyc (approved):
LGTM

PR #405, opened 2024-08-28
Add text_pair_similarity_filter

Qirui-jiao:
Calculate the CLIP text features for text pairs, and filter the samples based on their cosine similarity.
Hyperparameters:

- min_score – the min similarity score of filter range.
- max_score – the max similarity score of filter range.
- any_or_all – keep this sample when any/all text pairs meet the filter condition.


HYLcool:

PR #410, opened 2024-09-01
Use analyzer instead of analyser to maintain consistency

garyzhang99:
As the title says, fixing typo, use analyzer.yaml instead of analyser.yaml to maintain consistency.


drcege (approved):
LGTM

PR #421, opened 2024-09-09
fix: missing args in load_formatter of Analyzer

zhijianma:
As title says.


drcege:
Perhaps we should consider enforcing keyword-only parameters (separated by `*`) in many function signatures to avoid misuse.

yxdyc:
> Perhaps we should consider enforcing keyword-only parameters (separated by `*`) in many function signatures to avoid misuse.

Agree. Especially for the support of our new feature, DJ as a Service. We can add a new TODO item/issue for this problem.

yxdyc (approved):

PR #422, opened 2024-09-09
use pydantic types

drcege:
`Tuple[str]` or `Tuple[int]` was used incorrectly. They indicate that the tuple can only contain a single element, which differs from `List[str]` or `List[int]`, which are variable-length sequences by default. After the modification:
- All variable-length sequences are standardized to use `List`, which can be correctly recognized and converted by Pydantic from compatible types like `tuple`, `list`, `set`, `frozenset`, `deque`, etc. [ref](https://docs.pydantic.dev/2.0/usage/types/list_types/)
- Use `Tuple` only when the length of the sequence is fixed, such as a pair of coordinates that always has a length of 2.

Refactor all type annotations from `jsonargparse` to `Pydantic`.


zhijianma (approved):
LGTM

PR #423, opened 2024-09-09
Add new OP: image_tagging_mapper

HYLcool:
- Similar to video_tagging_from_frames_mapper
- Support user-specified tag field names for all tagging OPs
- fix the type bug in video_tagging_from_audio_mapper: force the type of result field to list of string.


data_juicer/ops/mapper/image_tagging_mapper.py
```diff
@@ -0,0 +1,80 @@
...
+        image_tags = []
+        for _, value in enumerate(loaded_image_keys):
+            image = images[value]
+
+            image_tensor = torch.unsqueeze(self.transform(image), dim=0).to(
+                next(model.parameters()).device)
+            with torch.no_grad():
+                tags, _ = model.generate_tag(image_tensor)
+
+            words = [word.strip() for tag in tags for word in tag.split('|')]
```
drcege, line 74: Since we process images one by one, `tags` should always be a list of length 1. It may be more straightforward to directly use `tags[0]` for consistency with `unsqueeze(dim=0)`.

data_juicer/ops/mapper/image_tagging_mapper.py
```diff
@@ -0,0 +1,80 @@
...
+        :param tag_field_name: the field name to store the tags. It's
+            "__dj__image_tags__" in default.
+        :param args: extra args
+        :param kwargs: extra args
+        """
+        super().__init__(*args, **kwargs)
+        self.model_key = prepare_model(
+            model_type='recognizeAnything',
+            pretrained_model_name_or_path='ram_plus_swin_large_14m.pth',
+            input_size=384)
```
drcege, line 44: Is it necessary to expose the `input_size` parameter?

data_juicer/ops/mapper/image_tagging_mapper.py
```diff
@@ -0,0 +1,80 @@
...
+@UNFORKABLE.register_module(OP_NAME)
+@OPERATORS.register_module(OP_NAME)
+@LOADED_IMAGES.register_module(OP_NAME)
+class ImageTaggingMapper(Mapper):
+    """Mapper to generate image tags.
+    """
+
+    _accelerator = 'cuda'
+
+    def __init__(self, tag_field_name=Fields.image_tags, *args, **kwargs):
```
drcege, line 32: Please add type annotations for all parameters, like `:str`

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
```diff
@@ -80,11 +88,11 @@ def process(self, sample, rank=None):
...
                                        sampling_rate=sr,
-                                       return_tensors='pt')
+                                       return_tensors='pt').to(model.device)
             with torch.no_grad():
                 logits = model(**inputs).logits
             predicted_tag_id = torch.argmax(logits, dim=-1).item()
             predicted_tag = model.config.id2label[predicted_tag_id]
             video_audio_tags.append(predicted_tag)
-        sample[Fields.video_audio_tags] = video_audio_tags
+        sample[self.tag_field_name] = np.array(video_audio_tags, dtype=np.str_)
```
drcege, line 97: Do you think it's necessary to unify the format of `tags` stats to match the other two OPs, using a double list?

tests/ops/mapper/test_image_tagging_mapper.py
```diff
@@ -0,0 +1,136 @@
...
+@SKIPPED_TESTS.register_module()
+class ImageTaggingMapperTest(DataJuicerTestCaseBase):
+    data_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..',
+                             'data')
+    img1_path = os.path.join(data_path, 'img1.png')
+    img2_path = os.path.join(data_path, 'img2.jpg')
+    img3_path = os.path.join(data_path, 'img3.jpg')
+
+    def _run_image_tagging_mapper(self,
+                                              op,
```
drcege, line 20: Indentation is a bit strange

configs/config_all.yaml
```diff
@@ -122,6 +122,8 @@ process:
       cv_classifier: ''                                       # OpenCV classifier path for face detection. By default, we will use 'haarcascade_frontalface_alt.xml'.
       blur_type: 'gaussian'                                   # type of blur kernel, including ['mean', 'box', 'gaussian']
       radius: 2                                               # radius of blur kernel
+  - image_tagging_mapper:                                   # Mapper to generate image tags.
+      tag_field_name: '__dj__image_tags__'                    # the field name to store the tags. It's "__dj__image_tags__" in default.
```
drcege, line 126: Just a thought: many built-in `constants` mix single and double underscores, which may be confusing. Is it necessary to standardize them to single underscores?

data_juicer/ops/mapper/image_tagging_mapper.py
HYLcool: Yes, it is.👍🏻

data_juicer/ops/mapper/image_tagging_mapper.py
HYLcool: I don't think it's necessary. The valid input image sizes are either 384 or 224 for this model. The default size 384 works well for most cases.

data_juicer/ops/mapper/image_tagging_mapper.py
HYLcool: Done.

tests/ops/mapper/test_image_tagging_mapper.py
HYLcool: Fixed.

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
HYLcool: video_tagging_from_audio_mapper will output a single string tag for each video, thus the tags should be: ["tag1", "tag2", ...]

However, the other two tagging OPs using ram library will output a list of string tags for each image or video. The tags are [["tag1-1", "tag1-2", ...], ["tag2-1", "tag2-2", ...], ...]. So they actually have different formats of `tags`.

For your notice, I unifiy the format of the other OPs to string list as well.

HYLcool: video_tagging_from_audio_mapper will output a single string tag for each video, thus the tags should be: ["tag1", "tag2", ...]

However, the other two tagging OPs using ram library will output a list of string tags for each image or video. The tags are [["tag1-1", "tag1-2", ...], ["tag2-1", "tag2-2", ...], ...]. So they actually have different formats of `tags`.

For your notice, I unifiy the format of the other OPs to string list as well.

configs/config_all.yaml
HYLcool: The rule is kinda simple:
1. special word "dj" should be surrounded by "__" (two underscores)
2. the whole field name should be surrounded by "__" (two underscores)
3. words in the field name can be separated by "_" (one underscore)

I think it's better to keep still because this change will make it incompatible with the processed datasets before.

data_juicer/ops/filter/video_tagging_from_frames_filter.py
```diff
@@ -37,6 +37,7 @@ def __init__(self,
                  contain: str = 'any',
                  frame_sampling_method: str = 'all_keyframes',
                  frame_num: PositiveInt = 3,
+                 tag_field_name=Fields.video_frame_tags,
```
drcege, line 40: missing annotation

data_juicer/ops/mapper/image_tagging_mapper.py
```diff
@@ -0,0 +1,85 @@
...
+        self.tag_field_name = tag_field_name
+
+    def process(self, sample, rank=None, context=False):
+        # check if it's generated already
+        if self.tag_field_name in sample:
+            return sample
+
+        # there is no image in this sample
+        if self.image_key not in sample or not sample[self.image_key]:
+            sample[self.tag_field_name] = np.array([[]], dtype=np.str_)
```
drcege, line 61: This produces an empty numpy array, but `image_tags` in line 84 is a python list?

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
drcege: Alright, the current `tagging` from the audio operation is essentially `classification` and can generally default to outputting only one label. 

My previous concern was that if `tagging` potentially needs to support multi-label output, unifying it to the same nested list would be more consistent. However, after reviewing several audio models, I found that most models are still primarily focused on classification tasks, so we can maintain the current implementation. If a true speech tagging (multi-label) model arises in the future, we can discuss adjustments then.

data_juicer/ops/mapper/image_tagging_mapper.py
HYLcool: Yes. The list in line 84 will be correctly converted to list[list[str]] by datasets. Here if we don't give the empty list an explicit type definition, it will be converted to null and can not be aligned with list[list[str]]

data_juicer/ops/filter/video_tagging_from_frames_filter.py
HYLcool: Done.

drcege (approved):

PR #429, opened 2024-09-13
Feat/dj adapter

HYLcool:
Support basic Adapter and Monitor:
- monitor resource utilization when running specified functions
- probe small data batch, monitor the resource utilization of each OP, and decide the batch size for each OP

Notice: datasets need to be updated to >= 2.19.0

- [ ] TODO: set batch size to 1 for all OPs before probing
- [ ] Wait for batch size support

### DIFF
--- a/data_juicer/core/adapter.py
+++ b/data_juicer/core/adapter.py
@@ -1,0 +1,81 @@
+from data_juicer.core.monitor import Monitor
+
+
+class Adapter:
+
+    def __init__(self):
+        self.current_resources = self.detect_current_resources()
+
+    @staticmethod
+    def execute_and_probe(dataset, operators, sample_interval=0.5):
+        """
+        Process the input dataset and probe related information for each OP in
+        the specified operator list.
+
+        For now, we support the following targets to probe:
+        "resource": resource utilization for each OP.
+        "speed": average processing speed for each OP.
+
+        The probe result is a list and each item in the list is the probe
+        result for each OP.
+        """
+        if operators is None or len(operators) == 0:
+            return []
+
+        # number of test samples
+        sample_num = len(dataset)
+        # resource utilization list
+        resource_util = []
+        # probe for each OP
+        for op in operators:
+
+            # run single op and monitor the resource utilization
+            dataset, resource_util_per_op = Monitor.monitor_func(
+                op.run, args=(dataset, ), sample_interval=sample_interval)
+
+            # calculate speed
+            resource_util_per_op[
+                'speed'] = sample_num / resource_util_per_op['time']
+            resource_util.append(resource_util_per_op)
+
+        return resource_util
+
+    def probe_small_batches(self, dataset, operators):
+        """
+        Perform small batch pre-execution to probe available resources,
+        current load and estimated OP speed, returning load factors and speed
+        ranks for each OP.
+
+        :param dataset: The dataset to pre-execute small batches on
+        :param operators: The OP list to be pre-execution and probe
+        :return: A list of probe results for each OP.
+        """
+        # 假设这个函数执行一小部分数据以检测负载
+        print('Pre-executing small batches to detect system load...')
+        # 这里可以添加具体的逻辑来预执行小批量
+        # 模拟的负载因子（可以根据实际情况计算）
+
+        # 例如，在负载过高的情况下返回较小的值
+        load_factor = self.available_resources['load']  # 返回当前负载
+        print(f'Detected load factor: {load_factor}')
+        return load_factor
+
+    @staticmethod
+    def batch_split(dataset, config, load_factor=None):
+        """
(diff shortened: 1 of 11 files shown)


data_juicer/core/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+                op.run, args=(dataset, ), sample_interval=sample_interval)
+
+            # calculate speed
+            resource_util_per_op[
+                'speed'] = sample_num / resource_util_per_op['time']
+            resource_util_list.append(resource_util_per_op)
+
+        return resource_util_list
+
+    def workloads_adapt(self, dataset, operators):
```
drcege, line 45: Looks like this is the main entrypoint; best to organize methods based on the call flow.

The name is somewhat unremarkable; might be `schedule/adapt_workloads/balance_workloads` to align with other naming styles.

data_juicer/core/adapter.py
```diff
@@ -0,0 +1,137 @@
+from data_juicer.core.monitor import Monitor
+
+
+class Adapter:
+
+    MAX_BATCH_SIZE = 10000
+
+    def __init__(self, cfg):
```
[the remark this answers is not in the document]
HYLcool: Done.

data_juicer/core/adapter.py
HYLcool: Right, `adapt_workloads` seems better~

data_juicer/core/monitor.py
```diff
@@ -0,0 +1,214 @@
...
+        Detect the resource utilization of all distributed nodes.
+        """
+        # TODO
+        raise NotImplementedError
+
+    @staticmethod
+    def monitor_current_resources():
+        """
+        Detect the resource utilization of the current environment/machine.
+        All data of "util." is in percent. All data of "mem." is in MB.
```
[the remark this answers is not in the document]
HYLcool: Yes. They were in percent before but changed to ratio later. Fixed.

data_juicer/core/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+    def take_batch(dataset, config):
+        """
+        Split the dataset into batches based on configuration and load factor.
+
+        :param dataset: The dataset to be split
+        :param config: Configuration settings, including batch size
+        :return: An iterator of batches
+        """
+        # get initial batch size
+        batch_size = config.get('batch_size', 1000)
```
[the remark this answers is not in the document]
HYLcool: To align HF datasets library, we can set the default batch size to 1000. Here I changed it to DEFAULT_MAX_BATCH_SIZE from datasets.config, which can also be configurable.

data_juicer/core/adapter.py
```diff
@@ -0,0 +1,137 @@
...
+
+class Adapter:
+
+    MAX_BATCH_SIZE = 10000
+
+    def __init__(self, cfg):
+        self.cfg = cfg
+        self.idle_resources = Monitor.monitor_current_resources()
+
+    @staticmethod
```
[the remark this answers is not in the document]
HYLcool: Done.

data_juicer/core/adapter.py
```diff
@@ -0,0 +1,147 @@
...
+
+        for item in load_analysis_res:
+            max_util = 1e-5
+            max_key = min(left_utils.items(), key=lambda it: it[1])[0]
+            analysis_res = item['resource_analysis']
+            for key in analysis_res:
+                if 'util.' not in key or 'GPU' in key:
+                    continue
+                used_util = max(
+                    0, analysis_res[key]['max'] - self.idle_resources[key])
```
yxdyc, line 138: to discuss: using `max` or `avg`? As `max` is sensitive to the outlier within the probed small batch, we may have high probability to under-leverage the hardware resources.

data_juicer/core/adapter.py
```diff
@@ -0,0 +1,147 @@
...
+        :return: A list of probe results for each OP.
+        """
+        # take a small batch
+        data_batch = self.take_batch(dataset, self.cfg)
+        # process and monitor the resource utilization
+        resource_util_list = self.execute_and_probe(data_batch, operators)
+        # analyze resource utilization
+        analysis_res = Monitor.analyze_resource_util_list(resource_util_list)
+
+        return analysis_res, len(data_batch)
```
yxdyc, line 110: In docstring "return: A list of probe results for each OP." --> "..., and the length of probed data batch

HYLcool: done.

data_juicer/core/adapter.py
yxdyc: to be tested with more experiments and will incorporated in following PRs

yxdyc (approved):

PR #435, opened 2024-09-24
Refine batch op branch

BeachWang:
1. Change default op batch size from 1 to 1000.
2. Change list(map()) to map() for filter OPs and keep origin coding for mapper OPs.
3. Make sure that dataset is a NestedDataset instance in `run` function.
NOTE: It does not make sure dataset to be NestedDataset instance when directly calling `process` function in Deduplicator and Selector OPs!


HYLcool (approved):
LGTM

PR #459, opened 2024-10-22
use chat_template

drcege:


drcege:
Already integrated in #454

PR #464, opened 2024-10-25
Probe-based OP Fusion & Reordering

HYLcool:
As the title says.

Also supports adaptive batch size.


configs/config_all.yaml
```diff
@@ -21,9 +21,11 @@ open_tracer: false                                          # whether to open th
 op_list_to_trace: []                                        # only ops in this list will be traced by tracer. If it's empty, all ops will be traced. Only available when tracer is opened.
 trace_num: 10                                               # number of samples to show the differences between datasets before and after each op. Only available when tracer is opened.
 op_fusion: false                                            # whether to fuse operators that share the same intermediate variables automatically. Op fusion might reduce the memory requirements slightly but speed up the whole process.
+fusion_strategy: 'greedy'                                   # OP fusion strategy. Support ['greedy', 'probe'] now. 'greedy' means keep the basic OP order and put the fused OP to the last of each fused OP group. 'probe' means Data-Juicer will probe the running speed for each OP at the beginning and reorder the OPs and fused OPs according to their probed speed (fast to slow). It's 'greedy' in default.
```
yxdyc, line 24: Changing default value into "probe"?

yxdyc (approved):

configs/config_all.yaml
HYLcool: Done

PR #479, opened 2024-11-08
[Ready] support API retry

drcege:
- Use OpenAI's `client.post` to be compatibility with OpenAI, Dashscope, etc., while allowing for customizable API endpoints.
- All parameters supported by OpenAI are accepted, such as `timeout`, `max_retries`, etc., which can be passed through the `model_params` parameter of the operator.
- Add mocked API unit tests.


BeachWang (approved):
LGTM

Issue #412, opened 2024-09-02
[Feat] Support `PythonCodesOperator` and `BashCodesOperator` that wraps an existing python file, or some code snippets to be executed, such as the existing DJ tools.


yxdyc:
### Search before continuing 先搜索，再继续

- [X] I have searched the Data-Juicer issues and found no similar feature requests. 我已经搜索了 Data-Juicer 的 issue 列表但是没有发现类似的功能需求。

### Description 描述

Often, users require the integration of specific Data-Juicer tools, custom functionalities encapsulated within some `helper_func.py`, or some short Python scripts, such as a few lambda functions. These may not warrant the creation of a dedicated Operator due to the additional overhead involved, including subclassing, documentation, and unit testing.

To enhance flexibility and cater to diverse user needs, introducing some OPs that seamlessly incorporates existing Python files or executes code snippets would be beneficial. This approach enables users to enrich their data recipe configurations with a wider array of tools and custom code, which can be managed through a streamlined `PythonCodesOperator` and `BashCodesOperator` mechanism.

### Are you willing to submit a PR for this feature? 您是否乐意为此功能提交一个 PR？

- [X] Yes I'd like to help by submitting a PR! 是的！我愿意提供帮助并提交一个PR！

yxdyc:
WIP: add basic abstraction and implementation for `ScriptOP` and some leaf classes such as `PythonCodesOperator` and `BashCodesOperator`, @drcege @BeachWang

HYLcool:
This feature request has been implemented in #492 and #493 by @drcege and @BeachWang .

PR #480, opened 2024-11-11
op/command_mapper

drcege:
Added the prototype implementation of `CommandMapper`, which corresponds to the `ScriptOP` raised in issue #412. The following issues need discussion:

- Avoid creating subclasses like `PythonCodesOperator` or `BashCodesOperator`, as this hinders extensibility. Consider that if users want to support more shell types like `zsh`, or integrate with other programming languages or third-party tools, they would have to add new classes at the code level, which is cumbersome and unnecessary. With a single `CommandMapper`, we can maximize support for arbitrary extensions without modifying the source code.

- Operator naming. Currently, the operator is implemented as a mapper; therefore, the `Mapper` suffix is used instead of the original `ScriptOP`. Other candidate names include `ScriptMapper`, `CodeMapper`, `ExternalXXXMapper`, etc. I have opted for `CommandMapper` because it is not limited to just code or scripts; it can actually execute any command, such as pre-compiled binary files.

Everyone is welcome to provide feedback on the above issues, and I will add relevant documentation after reaching a consensus.

### DIFF
--- a/data_juicer/ops/mapper/__init__.py
+++ b/data_juicer/ops/mapper/__init__.py
@@ -8,6 +8,7 @@ from .clean_email_mapper import CleanEmailMapper
 from .clean_html_mapper import CleanHtmlMapper
 from .clean_ip_mapper import CleanIpMapper
 from .clean_links_mapper import CleanLinksMapper
+from .command_mapper import CommandMapper
 from .expand_macro_mapper import ExpandMacroMapper
 from .fix_unicode_mapper import FixUnicodeMapper
 from .generate_qa_from_examples_mapper import GenerateQAFromExamplesMapper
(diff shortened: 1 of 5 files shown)


yxdyc:
Good prototype and thoughts on the PR. I concur with the first point regarding avoiding the creation of multiple subclasses for handling different shell environments, programming languages, or tools. This approach does indeed support greater extensibility and reduces code maintenance burdens.

However, I think that abstracting a ScriptOP base class can still be beneficial, especially for purposes related to security and fault tolerance. For example, a unified management interface could be used to restrict high-risk system commands like `rm`, or to consistently check command return statuses. Additionally, for improved efficiency and flexibility, we could consider maintaining two subclasses with the option for `batched=False/True`, allowing users to choose the most suitable approach for their needs. For instance:
```
CommandMapper (command: str = '', batched=False, repair=True)
CommandFilter (command: str = '', batched=False, repair=True)
```

drcege:
Moved to #492

Issue #440, opened 2024-09-29
[Bug]: KeyError: 'resource'


luckystar1992:
### Before Reporting 报告之前

- [X] I have pulled the latest code of main branch to run again and the bug still existed. 我已经拉取了主分支上最新的代码，重新运行之后，问题仍不能解决。

- [X] I have read the [README](https://github.com/alibaba/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process. (Otherwise, we recommend that you can ask a question using the Question template) 我已经仔细阅读了 [README](https://github.com/alibaba/data-juicer/blob/main/README_ZH.md) 上的操作指引，并且在安装过程中没有错误发生。（否则，我们建议您使用Question模板向我们进行提问）

### Search before reporting 先搜索，再报告

- [X] I have searched the Data-Juicer [issues](https://github.com/alibaba/data-juicer/issues) and found no similar bugs. 我已经在 [issue列表](https://github.com/alibaba/data-juicer/issues) 中搜索但是没有发现类似的bug报告。

### OS 系统

MacOS15.0

### Installation Method 安装方式

source

### Data-Juicer Version Data-Juicer版本

0.2.0

### Python Version Python版本

3.10

### Describe the bug 描述这个bug

最新的main分之的代码，运行算子的时候，出现以下错误信息：
```
Traceback (most recent call last):
  File "/Users/zyc/code/data-juicer/data_juicer/core/data.py", line 199, in process
    dataset, resource_util_per_op = Monitor.monitor_func(
                                    ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zyc/code/data-juicer/data_juicer/core/monitor.py", line 210, in monitor_func
    resource_util_dict['resource'] = mdict['resource']
                                     ~~~~~^^^^^^^^^^^^
  File "<string>", line 2, in __getitem__
  File "/Users/zyc/miniconda3/lib/python3.11/multiprocessing/managers.py", line 837, in _callmethod

    raise convert_to_error(kind, result)
KeyError: 'resource'
```

而在之前的代码中不会抱这个错误。试了几个算子都是这样。

### To Reproduce 如何复现

python process_data.py --config ../configs/demo/process_demo.yaml  

配置文件如下：

`# Process config example for dataset

# global parameters
project_name: 'demo-process'
dataset_path: '/Users/zyc/code/data-juicer/demos/data/demo-dataset-chatml.jsonl'
np: 1  # number of subprocess to process your dataset
[...]

PR #483, opened 2024-11-12
Performance benchmark

HYLcool:
Performance benchmark for 4 modalities. Reports will be uploaded to internal wandb server.

TODO:

- [x] record resource utilization info after the PR #464 is merged.

Others:
- Add an argument to control whether to open `Monitor` for data processing. It's True by default. (to resolve issue #440 )
- The analyzed results of monitored resource utilization are added into the monitor report file as well now.
- For the mp start method of monitor, set it to "spawn" for Windows systems and "fork" for others. (to resolve issue #477 )
- Add backup link for ram model.
- Update transformers version to >=4.47.0 to avoid "shape mismatch" bug from older version 4.46.3. (ref https://github.com/huggingface/transformers/issues/34990)


BeachWang (approved):
LGTM

PR #500, opened 2024-11-27
add grouper and aggregator op for system_prompt

BeachWang:


yxdyc (approved):
LGTM. In the future, we can extend the DEFAULT_SYSTEM/XXX_PROMPT to include both Chinese and English versions for broader accessibility and applicability.

PR #503, opened 2024-11-29
Quick fix for some minor problems

HYLcool:
* remove str conversion for fps para of add_stream func
* add requires from librosa to avoid lazy_loader failure during multiprocessing


BeachWang (approved):
LGTM

PR #505, opened 2024-12-05
Registe all other formatters

jackylee-ch:
We are using `RemoteFormatter` in our environment, and found it is not registered.

### DIFF
--- a/data_juicer/format/formatter.py
+++ b/data_juicer/format/formatter.py
@@ -19,6 +19,7 @@ class BaseFormatter:
         raise NotImplementedError
 
 
+@FORMATTERS.register_module()
 class LocalFormatter(BaseFormatter):
     """The class is used to load a dataset from local files or local
     directory."""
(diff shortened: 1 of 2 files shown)


HYLcool:
Hi @jackylee-ch , thanks for your contribution!

`FORMATTER` registry is for **local** formatters to guide `LocalFormatter` to the **specific** `Formatter` for the formats of files in the local datasets, e.g. JsonFormatter for '.json', '.jsonl', '.jsonl.zst' files. Therefore, we don't need to register `LocalFormatter` because it's still an **abstract** class for those specific `Formatter`. `RemoteFormatter` is used to handle **remote** datasets on the HF Hub so we don't register it as well. As for `MixtureFormatter`, it's used to mix several datasets and it's not tied to any specific format.

jackylee-ch:
Got it. It would automically use RemoteFormatter if it is not an absolute path.

PR #506, opened 2024-12-06
Patch for Perf Bench

HYLcool:
* refine perf bench workflow
* fix wrong var in sphinx docs
* enable several unit tests that were skipped due to dependency problems.


BeachWang (approved):
LGTM

PR #511, opened 2024-12-11
support ray actor

Cathy0908:


data_juicer/ops/base_op.py
```diff
@@ -172,6 +174,9 @@ def __init__(self, *args, **kwargs):
     def is_batched_op(self):
         return self._batched_op
 
+    def use_ray_actor(self):
+        return self._ray_mode == 'actor'
```
yxdyc, line 178: Perhaps we do not need a new field like `_ray_mode`, as it requires users to be aware of Ray usage and one more step when developing new OPs. Instead of, we can re-use existing field  `_accelerator = 'cuda'`, since the root issue involves large model parameters and high VRAM, which are typically associated with the use of GPUs.

HYLcool (approved):
LGTM

PR #512, opened 2024-12-13
[Feature] add auto mode for analyzer

HYLcool:
```bash
dj-analyze --auto --dataset_path xx.jsonl [--auto_num 1000]
```

- add auto mode for **Analyzer**, so users can:
  - analyze a small part of input dataset quickly and avoid writing a specific recipe
  - with all Filters that produce stats
  - change the number of samples to be analyzed in auto mode using `--auto_num` arg.
- `--auto` and `--config` are mutually exclusive, meaning users can provide a specific recipe or analyze in auto mode.

Others:
- support drawing word cloud images for string stats in addition to histograms.
- set default `mem_required` for all model-based OPs.
- limit wandb version to "<=0.19.0" because the latest 0.19.1 could cause query exception

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,11 @@ python tools/analyze_data.py --config configs/demo/analyzer.yaml
 
 # use command line tool
 dj-analyze --config configs/demo/analyzer.yaml
+
+# you can also use auto mode to avoid writing a recipe. It will analyze a small
+# part (e.g. 1000 samples, specified by argument `auto_num`) of your dataset 
+# with all Filters that produce stats.
+dj-analyze --auto --dataset_path xx.jsonl [--auto_num 1000]
 ```
 
 - **Note:** Analyzer only compute stats of Filter ops. So extra Mapper or Deduplicator ops will be ignored in the analysis process.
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -317,6 +317,11 @@ python tools/analyze_data.py --config configs/demo/analyzer.yaml
...
 # use command line tool
 dj-analyze --config configs/demo/analyzer.yaml
+
+# you can also use auto mode to avoid writing a recipe. It will analyze a small
+# part (e.g. 1000 samples, specified by argument `auto_num`) of your dataset 
+# with all Filters that produce stats.
+dj-analyze --auto --dataset_path xx.jsonl [--auto_num 1000]
 ```
 
 - **Note:** Analyzer only compute stats of Filter ops. So extra Mapper or Deduplicator ops will be ignored in the analysis process.
```
yxdyc, line 327: For all OPs that can generates dj_stats or meta.xx_dj_tags

data_juicer/ops/mapper/video_captioning_from_summarizer_mapper.py
```diff
@@ -81,6 +81,7 @@ def __init__(self,
         :param args: extra args
         :param kwargs: extra args
         """
+        kwargs.setdefault('mem_required', '40GB')
```
yxdyc, line 84: Perhaps we may add new Fields like `mem_required_quant_8` for these large-model-based OPs

yxdyc (approved):
LGTM

PR #518, opened 2024-12-24
Dev/manage meta

BeachWang:
Have organized and consolidated the generated metadata uniformly.
1. Meta data produced by mapper OPs is stored in `Fields.meta` field. Corresponding keys are list in `MetaKeys`.
2. For convenience of split a batch sample, meta data produced by aggregator OPs is stored in `Fields.agg` field. Corresponding keys are list in `AggKeys`. Here, we add a default `Fields.agg` field to all samples in all aggregator operators, without adding a decorator marker `TAGGING_OPS` like we do with mappers.
3. Unify the interface for generating meta operators, allowing modification of meta keys. If a meta key already exists, it should simply return without making any changes.
4. Add corresponding unit test.
5. Adjust the role playing demo according to the changes.
6. Remove `nested_set` function.


yxdyc (commented):
How about removing `Fields.agg` and `AggKeys`? Instead, we can unify them into `Fields.meta` and `MetaKeys`. This makes sense because they are implemented in a very similar way, and conceptually, `agg` is still a type of `meta`. 

Additionally, currently, we only have two types of `AggKeys`, and keeping both `agg` and `meta` may increase the cognitive load for users and introduce inconsistencies. Our 2.0 paper mentions that we have three main parts of the fields: 1. data payload, 2. stats, 3. meta. Unifying `agg` into `meta` would align better with this structure.

Others LGTM.

yxdyc (approved):

PR #521, opened 2024-12-25
Fix operators doc link for aggregators

jackylee-ch:
The link for aggregators tests in `Operators.md` and `Operators_ZH.md` will lead to `page not found`. This pr would rename the directory name so that the links can find the test files.


yxdyc (approved):
LGTM, thanks for your contribution~

PR #523, opened 2024-12-26
docs for distributed processing

HYLcool:
As the title says.


yxdyc:
I made some modifications in [this commit](https://github.com/modelscope/data-juicer/pull/523/commits/1f131df1d0421f51397ddd57ecbd321afd2dceec). Please @HYLcool, double-check and translate it into the Chinese version.

Additionally, @pan-x-c and @chenyushuo, do we need to add any other information or examples in the quick-start part?

docs/Distributed.md
```diff
@@ -0,0 +1,104 @@
...
+- The exception is the deduplication operators, which are challenging to scale in standalone mode. We provide these operators with the prefix [`ray_xx_deduplication`](../data_juicer/ops/deduplicator/).
+
+### Subset Splitting
+
+When dealing with tens of thousands of nodes but only a few dataset files, Ray would split the dataset files according to available resources and distribute the blocks across all nodes, incurring huge network communication costs and reduces CPU utilization. For more details, see [Ray's autodetect_parallelism](https://github.com/ray-project/ray/blob/2dbd08a46f7f08ea614d8dd20fd0bca5682a3078/python/ray/data/_internal/util.py#L201-L205) and [tuning output blocks for Ray](https://docs.ray.io/en/latest/data/performance-tips.html#tuning-output-blocks-for-read).
+
+To optimize performance, we automatically split the original dataset into smaller files in advance, considering the features of Arrow and Ray. By default, the single file size is set to 128MB if it results in the number of sub-files being larger than five times the total number of CPU cores in the cluster.
```
pan-x-c, line 27: The description is incorrect. Splitting is no longer required with streaming loading.

HYLcool:
> I made some modifications in [this commit](https://github.com/modelscope/data-juicer/pull/523/commits/1f131df1d0421f51397ddd57ecbd321afd2dceec). Please @HYLcool, double-check and translate it into the Chinese version.

The Chinese version is ready.

docs/Distributed.md
HYLcool: I don't get it because this subsection did not involve streaming loading.

yxdyc: If I understand correctly, the Ray's `autodetect_parallelism` func is not aware of parameters related to streaming loading, neither in this function or its internally called functions like `datasource_or_legacy_reader.estimate_inmemory_data_size()` and `_estimate_avail_cpus`. 

As a result, if the condition *tens of thousands of nodes but only a few dataset files* holds, an in-efficient execution plan could still be generated in `get_read_tasks`, although users set streaming-read options.

yxdyc: BTW, we can explicitly provide this feature and the below streaming-load feature after zhizhu's refactoring of auto_formater & data_loaders. E.g.,  DJ helps users to automatically split the dataset into suitable sub-files to avoid in-efficient Ray plan generation, as well as set streaming-related options to avoid OOM issue.

yxdyc (approved):
You can double-check the current version, which LGTM.

PR #536, opened 2025-01-09
fix save_ckpt bug

HYLcool:
- IndexError if the number of samples in the result dataset is less than the number of workers when saving dataset to disk


BeachWang (approved):
LGTM

PR #541, opened 2025-01-10
limit the generated qa num for each text in generate_qa_from_text_mapper

BeachWang:
As the title say.

### DIFF
--- a/tests/ops/mapper/test_generate_qa_from_text_mapper.py
+++ b/tests/ops/mapper/test_generate_qa_from_text_mapper.py
@@ -19,11 +19,13 @@ class GenerateQAFromTextMapperTest(DataJuicerTestCaseBase):
                 enable_vllm=False,
                 model_params=None,
                 sampling_params=None,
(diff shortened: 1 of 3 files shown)


tests/ops/mapper/test_generate_qa_from_text_mapper.py
```diff
@@ -19,11 +19,13 @@ def _run_op(self,
...
                 sampling_params=None,
-                num_proc=1):
+                num_proc=1,
+                max_num=None):
 
         op = GenerateQAFromTextMapper(enable_vllm=enable_vllm,
                                       model_params=model_params,
-                                      sampling_params=sampling_params)
+                                      sampling_params=sampling_params,
+                                      max_num=max_num)
```
HYLcool, line 28: Maybe we can add an assertion for this argument, e.g.
if max_num is not None:
    self.assertLessEqual(len(results), max_num)

HYLcool (approved):
Others LGTM

PR #545, opened 2025-01-14
Fix bug and add gif demo for role playing

BeachWang:
1. Fix typo in `system_prompt_generator.py`.
2. Fix init for extracted meta info since `dataset` assert the type in each column as same as the first sample so this field should not be empty in the first sample.
3. Add gif demo for role playing.


HYLcool (approved):
LGTM. @yxdyc double check.

PR #554, opened 2025-01-17
Update translator of OP doc building.

HYLcool:
- Use translator of alibaba, which is more stable and faster
  - previous: 7-8s
  - now: 2-3s


yxdyc (approved):
Modal Tags should be Model Tags. Others LGTM

PR #566, opened 2025-02-06
update the 2.0 paper link & the DaaR news

yxdyc:
as the title says. 

- The source codes of DaaR are being cleaned and added to [this branch](https://github.com/modelscope/data-juicer/tree/DaaR).
- The related new algorithm and features are being incorporated into the Sandbox (integrating OpenCompass & Swift) and OpZoo (as useful filter operations for post-tuning usages).

@lingzhq , we have already made the DJ data format compatible with Swift (see this [documentation](https://github.com/modelscope/data-juicer/blob/main/tools/fmt_conversion/post_tuning_dialog/README.md)). Feel free to discuss any other potential issues.


HYLcool (approved):
LGTM

PR #586, opened 2025-02-25
Minor opts for CI/CD

HYLcool:
- only keep py3.10 for doc publishment to avoid confliction.
- upload coverage reports in html after unit tests
- use primitive `@unittest.skip` instead of our `SKIPPED_TESTS`
- limit huggingface_hub to avoid undefined cache problem.

Others:
- make the new OP ImageRemoveBackgroundMapper stable
- avoid ModuleNotFound error for ray when installing minimal requirements by lazy-loading those ray-based deduplicators


yxdyc (approved):
LGTM

PR #590, opened 2025-02-25
fix: image_nsfw_filter.py memory leak. disable gradient

danielhjz:
when using cpu to run this op, memory continuously  rising.


HYLcool (approved):
Thanks for you contribution!

PR #592, opened 2025-02-26
fix ray dependency

BeachWang:
Remove ray dependency in basic environment.

### DIFF
--- a/data_juicer/ops/deduplicator/ray_basic_deduplicator.py
+++ b/data_juicer/ops/deduplicator/ray_basic_deduplicator.py
@@ -1,13 +1,12 @@
 from abc import ABC, abstractmethod
 
-import ray
-
 from data_juicer.utils.constant import HashKeys
 from data_juicer.utils.lazy_loader import LazyLoader
 
 from ..base_op import Filter
 
 redis = LazyLoader('redis', 'redis')
+ray = LazyLoader('ray', 'ray')
 
 MERSENNE_PRIME = (1 << 61) - 1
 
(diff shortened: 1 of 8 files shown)


yxdyc (approved):
LGTM

data_juicer/ops/deduplicator/ray_basic_deduplicator.py
```diff
@@ -1,13 +1,12 @@
...
 
-import ray
-
 from data_juicer.utils.constant import HashKeys
 from data_juicer.utils.lazy_loader import LazyLoader
 
 from ..base_op import Filter
 
 redis = LazyLoader('redis', 'redis')
+ray = LazyLoader('ray', 'ray')
```
HYLcool, line 9: If it's OK when the ray is immediately used below `@ray.remote`?

BeachWang: Oh, it will cause the ray auto installation immediately. So we need add ray to the basic dependency?

HYLcool: I think so. I tried lazy-load these ray-based OPs but it causes errors when running in ray mode (ref: ops/deduplicator/__init__.py in PR #586 , and corresponding failed unit test: https://github.com/modelscope/data-juicer/actions/runs/13539669463)

What I can come up with is the same as you said -- add ray to minimal requirements. Maybe you can modify it and help to restore ops/deduplicator/__init__.py

HYLcool: Nevermind, I modify this part in my latest PR #594 👌🏻

HYLcool: You can erase all lazy-load for ray in this PR.

PR #597, opened 2025-02-28
Convenient function call for getting op name to stats

BeachWang:
As the title says.


HYLcool (approved):

data_juicer/utils/constant.py
```diff
@@ -129,9 +129,23 @@ def __getattr__(cls, attr):
...
+            tmp_dj_cfg = copy.deepcopy(dj_cfg)
+            tmp_dj_cfg.use_cache = False
+            tmp_dj_cfg.use_checkpoint = False
+
+            from data_juicer.core import Analyzer
+            tmp_analyzer = Analyzer(tmp_dj_cfg)
+
+            from data_juicer.core.data import NestedDataset
+            dataset = NestedDataset.from_dict(
+                {k: [v]
```
HYLcool, line 145: Why do we need to convert v to [v]?

BeachWang: We only catch the first sample to construct the dataset for mini test execution.

HYLcool: Gotcha, maybe `dataset.take(1)` is what you need.

yxdyc (approved):

PR #598, opened 2025-02-28
Initial version of unit test only on diff

HYLcool:
* split unit-test to regression and partial versions
* set the trigger time of regression test to 7:00 on every Friday in Beijing time (finished nearly at 9:00)
* replace the unit test before with the partial test, which only test on the corresponding unit tests of changed files. If there is no corresponding unit tests, run all tests.


BeachWang (approved):
LGTM

PR #613, opened 2025-03-11
Add API service doc

BeachWang:
1. Add API service doc.
2. Our test fail in old version of `fastapi`. So I update it in env requirement.
3. Add more available function for OPs.
4. Add json dump parser for common list or dict params.


yxdyc (approved):
I updated some info for the homepage & data scoring, others LGTM.

PR #617, opened 2025-03-13
Prototype Implementation for HumanOps (annotation)

cyruszhang:
This feature is to add human annotation support for DJ. Included features:
- boilerplate code for supporting label studio powered human annotation ops
- a human preference annotation reference implementation is provided
- label studio service script; can start up local instance using docker or pip, whichever is available
- reference configs and data
- event driven and notification mixins framework for ops

still working on
- more efficient interaction with label studio SDK
- better logging with label studio localhost service
- notification details


cyruszhang:
ongoing work
- more efficient interaction with label studio SDK
- better logging with label studio localhost service
- notification details
- test case not working properly yet
- documentations

configs/config_all.yaml
```diff
@@ -950,3 +950,46 @@ process:
       try_num: 3                                              # The number of retry attempts when there is an API call error or output parsing error.
       model_params: {}                                        # Parameters for initializing the API model.
       sampling_params: {}                                     # Extra parameters passed to the API call. e.g {'temperature': 0.9, 'top_p': 0.95}
+
+# Notification configurations
```
cyruszhang, line 954: these haven't been ironed out yet.  might tweak later

data_juicer/ops/mapper/annotation/human_preference_annotation_mapper.py
```diff
@@ -0,0 +1,233 @@
...
+    LabelStudioAnnotationMapper
+
+from ...base_op import OPERATORS
+
+
+@OPERATORS.register_module('human_preference_annotation_mapper')
+class HumanPreferenceAnnotationMapper(LabelStudioAnnotationMapper):
+    """Operator for human preference annotation using Label Studio."""
+
+    DEFAULT_LABEL_CONFIG = """
```
cyruszhang, line 16: xml in code. could be a separate file if needed. supported via label_config_file param

data_juicer/ops/mapper/annotation/human_preference_annotation_mapper.py
```diff
@@ -0,0 +1,233 @@
...
+            sample: The original sample that was annotated
+
+        Returns:
+            Dict: The updated sample with preference results
+        """
+        # Extract the preference information
+        logger.debug(f'Processing annotation result: {annotation}')
+
+        preference = None
+        for item in annotation['result']:
```
cyruszhang, line 214: this can be tweaked. right now, we will have {'result': 'answer1'} as the result data field

cyruszhang:
getting below message:
2025-03-14 12:32:51 | WARNING  | data_juicer.core.tracer:63 - Datasets before and after op [human_preference_annotation_mapper] are all the same. Thus no comparison results would be generated.

tracer only checks text_key field. we are actually adding more data. what is the best resolution? should we pack all the data into text_key as json dump? @HYLcool @yxdyc

data_juicer/ops/mapper/annotation/human_preference_annotation_mapper.py
yxdyc: When end-to-end debugging this feature, @HYLcool we can adjust accordingly based on the fields of experience/RftDataset

yxdyc:
> getting below message: 2025-03-14 12:32:51 | WARNING | data_juicer.core.tracer:63 - Datasets before and after op [human_preference_annotation_mapper] are all the same. Thus no comparison results would be generated.
> 
> tracer only checks text_key field. we are actually adding more data. what is the best resolution? should we pack all the data into text_key as json dump? @HYLcool @yxdyc

After discussion, we agree to packing *all the data into text_key as json dump*, and will make a quick PR to support this feature @HYLcool

data_juicer/ops/mapper/annotation/human_preference_annotation_mapper.py
cyruszhang: Can you elaborate? A sample would be great

yxdyc (approved):
LGTM. In order to quickly integrate and debug, this version is merged first. The HumanOP feature is being continuously improved, within another practical human-in-the-loop use case.

PR #645, opened 2025-04-18
Human op enhancement

HYLcool:
- change the logic of human op to store the annotation results in chosen/rejected fields instead of only one result field.
- fix bugs of reading mm dataset


cyruszhang (approved):
LGTM

PR #650, opened 2025-04-23
Fix docker building failure

HYLcool:
As the title says.


Dockerfile
```diff
@@ -3,19 +3,21 @@
 
 FROM nvidia/cuda:12.4.1-cudnn-devel-ubuntu22.04
 
+# change to aliyun source
+RUN sed -i 's/archive.ubuntu.com/mirrors.aliyun.com/g' /etc/apt/sources.list \
```
cyruszhang, line 7: aliyun mirrors always more reliable?

tests/ops/filter/test_llm_quality_score_filter.py
```diff
@@ -47,7 +47,7 @@ def test_default_case(self):
             "Cats are domesticated animals known for their agility, intelligence, and independent nature. Research shows that they spend approximately 70% of their lives sleeping, which helps conserve energy for hunting. Unlike dogs, cats are obligate carnivores, meaning their diet must consist primarily of meat to meet nutritional needs."
         }]
         dataset = Dataset.from_list(ds_list)
-        op = LLMQualityScoreFilter(api_or_hf_model=self.api_or_hf_model)
+        op = LLMQualityScoreFilter(min_score=0.5, api_or_hf_model=self.api_or_hf_model)
```
cyruszhang, line 50: why are we defaulting to 0.0 and adding 0.5 in param? seems redundant; might as well leave the default to 0.5

Dockerfile
HYLcool: I'm not sure. But for the connection error we met before, it was resolved by this change. So maybe it's more reliable for our CI/CD machine.

tests/ops/filter/test_llm_quality_score_filter.py
HYLcool: These are changed for Trinity-RFT. But you're right, we should keep the default values as before, and modify them in the downstream apps.

yxdyc (commented):
LGTM

PR #658, opened 2025-04-30
Add Img-Diff recipes.

Qirui-jiao:
Add two ops and one recipe related to Img-Diff:
- imgdiff_difference_area_generator_mapper
- imgdiff_difference_caption_generator_mapper
- img-diff-recipe.yaml

PR #660, opened 2025-04-30
Better dependency management with uv & sub-groups

cyruszhang:
- introduce uv for better dependency management and resolution
- use pyproject.toml instead of setup.py; modern approach
- consolidate environemtns/*.requires.txt into pyproject.toml
- get rid of static mapping for lazy_loader logic; more dynamic dependency resolution 
- DJ-lite for minimal IO
- consolidate other requirements into scenario based groups 
- fix pre-commit related hiccups


pyproject.toml
```diff
@@ -0,0 +1,176 @@
...
+name = "py-data-juicer"
+dynamic = ["version"]
+description = "A One-Stop Data Processing System for Large Language Models."
+authors = [
+    {name = "SysML Team of Alibaba Tongyi Lab"}
+]
+readme = "README.md"
+license = {text = "Apache-2.0"}
+requires-python = ">=3.10"
+dependencies = [
```
cyruszhang, line 11: the default dependency has minimal IO related packages.

pyproject.toml
```diff
@@ -0,0 +1,176 @@
...
+    "jsonlines",  # JSONL handling
+    "zstandard",  # compression
+    "lz4",  # compression
+    "multiprocess==0.70.12",  # parallel processing
+    "dill==0.3.4",  # serialization
+    "psutil",  # system monitoring
+    "pydantic>=2.0",  # data validation
+]
+
+[project.optional-dependencies]
```
cyruszhang, line 28: we can ponder on how these groupings should work

vision/nlp/audio/generic/sandbox/distrbuted/dev

data_juicer/ops/__init__.py
```diff
@@ -1,6 +1,4 @@
-from . import aggregator, deduplicator, filter, grouper, mapper, selector
```
cyruszhang, line 1: yapf precommit complained

cyruszhang:
image
faster resolution with uv; much better parallelism with rust.

data_juicer/ops/__init__.py
cyruszhang: turns out the importing is very necessary to initialize the OPs registry. but since it's importing all the ops everytime you run pipelines, it can bring major delays; in local test, I am counting 50 seconds for the initialization . a better approach would be to check config file and decide what OPs to register

cyruszhang:
when running more local tests with a clean install, found some dependency missing issues
- recognize-anything is not properly initialized; test_video_tagging_from_frames_filter test_image_tagging_mapper would fail
- ffmpeg/ffmpeg-python is not properly initialized. test_audio_ffmpeg_wrapped_mapper test_video_ffmpeg_wrapped_mapper  

these tests are actually succeeding in github test workflows, because the dependencies are already properly setup, thus not from a clean install from scratch

HYLcool:
`samplerate` is implicitly imported by `librosa` in a lazy-loading way, which might cause an error. Maybe you need to test it from a clean installation. For example, only install `librosa` and do not install  `samplerate` manually, then try to run a recipe that contains audio-related OPs (audio_duration_filter/audio_nmf_snr_filter).

cyruszhang:
> `samplerate` is implicitly imported by `librosa` in a lazy-loading way, which might cause an error. Maybe you need to test it from a clean installation. For example, only install `librosa` and do not install `samplerate` manually, then try to run a recipe that contains audio-related OPs (audio_duration_filter/audio_nmf_snr_filter).

Did try to run these filter tests and they work fine. Resampy is used by default  instead of samplerate.

image
image
image
image

image

cyruszhang:
> > For now, the basic demo can not work when installing dj with default dependencies. I think we should add more necessary deps in the default dependency list (e.g. matplotlib & wordcloud required by analyzers)
> 
> Maybe you should add all dependencies in the original minimal requirements to the default dependencies, then decide which ones can be moved to other optional dependency lists by lazy-loading or some other ways through experiments from new clean environments.

good point. how we organize minimal requirements and other groupings should be discussed further

cyruszhang:
> when running more local tests with a clean install, found some dependency missing issues
> 
> * recognize-anything is not properly initialized; test_video_tagging_from_frames_filter test_image_tagging_mapper would fail
> * ffmpeg/ffmpeg-python is not properly initialized. test_audio_ffmpeg_wrapped_mapper test_video_ffmpeg_wrapped_mapper
> 
> these tests are actually succeeding in github test workflows, because the dependencies are already properly setup, thus not from a clean install from scratch

these issues has been resolved; from clean install, all work fine.
- proper module name and package name handling
- proper handling of direct github install and dependency resolution (with repo download and pip install)
- making sure ffmpeg-python opencv-python ram, so on and so forth, are properly referenced in LazyLoader

[f6d4aee9c41091bccc43468f938e17270a2b03f1
](https://github.com/modelscope/data-juicer/pull/660/commits/f6d4aee9c41091bccc43468f938e17270a2b03f1)
[42a1979c0c5a4619f4a9c9a24d310b778d94dcae](https://github.com/modelscope/data-juicer/pull/660/commits/42a1979c0c5a4619f4a9c9a24d310b778d94dcae)
[0fdfddc055d285ad33f3d89d0cdf25441fbb601a](https://github.com/modelscope/data-juicer/pull/660/commits/0fdfddc055d285ad33f3d89d0cdf25441fbb601a)

cyruszhang:
auto test didn't run somehow. manual run is here: https://github.com/modelscope/data-juicer/actions/runs/14916019849/job/41901689953

HYLcool:
> auto test didn't run somehow. manual run is here: https://github.com/modelscope/data-juicer/actions/runs/14916019849/job/41901689953

This failure is caused by lazy-loading `audiomentations`, which installs the `numpy>=2.0`. It can be resolved by adding it to the requirement list expilcitly (As PR #658 did).

data_juicer/ops/common/prompt2prompt_pipeline.py
```diff
@@ -1,12 +1,11 @@
...
 
-import numpy as np
-
 from data_juicer.utils.lazy_loader import LazyLoader
 
-diffusers = LazyLoader('diffusers', 'diffusers')
-torch = LazyLoader('torch', 'torch')
+diffusers = LazyLoader('diffusers')
+torch = LazyLoader('torch')
+np = LazyLoader('numpy')
```
HYLcool, line 8: numpy is in the minimal requirements, so there is no need to lazy load it.

data_juicer/ops/mapper/audio_add_gaussian_noise_mapper.py
```diff
@@ -11,7 +11,7 @@
 from ..base_op import OPERATORS, Mapper
 from ..op_fusion import LOADED_AUDIOS
 
-audiomentations = LazyLoader('audiomentations', 'audiomentations')
+audiomentations = LazyLoader('audiomentations')
```
HYLcool, line 14: Lazy-loading `audiomentations` might upgrade numpy to ">=2.0" version and cause some errors. Maybe we need to find some ways to limit the numpy version to the one expected in our toml file when lazy-loading it.

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
```diff
@@ -1,16 +1,14 @@
-import librosa
 import numpy as np
 
 from data_juicer.utils.constant import Fields, MetaKeys
-from data_juicer.utils.lazy_loader import AUTOINSTALL, LazyLoader
+from data_juicer.utils.lazy_loader import LazyLoader
 from data_juicer.utils.mm_utils import extract_audio_from_video
-from data_juicer.utils.model_utils import get_model, prepare_model
+from data_juicer.utils.model_utils import get_model, prepare_model, torch
```
HYLcool, line 6: Do we need to lazy-load torch as well?

data_juicer/ops/mapper/video_tagging_from_frames_mapper.py
```diff
@@ -4,18 +4,14 @@
 from pydantic import PositiveInt
 
 from data_juicer.utils.constant import Fields, MetaKeys
-from data_juicer.utils.lazy_loader import LazyLoader
 from data_juicer.utils.mm_utils import (close_video, extract_key_frames,
                                         extract_video_frames_uniformly,
                                         load_data_with_context, load_video)
-from data_juicer.utils.model_utils import get_model, prepare_model
+from data_juicer.utils.model_utils import get_model, prepare_model, ram, torch
```
HYLcool, line 10: Why the lazy-loadings of these two libs are removed? Same in `image_tagging_mapper`

cyruszhang: ram and torch in model_utils are already lazy loaded version

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
cyruszhang: torch is already lazy loaded in model_utils

data_juicer/ops/mapper/audio_add_gaussian_noise_mapper.py
cyruszhang: good point. when toml has the versioning, stick to it; if that cannot be satisfied, we should raise errors

data_juicer/ops/common/prompt2prompt_pipeline.py
cyruszhang: good catch

data_juicer/ops/mapper/video_tagging_from_audio_mapper.py
HYLcool: OK, I see.👌🏻

data_juicer/ops/mapper/audio_add_gaussian_noise_mapper.py
cyruszhang: logic added in newest commit https://github.com/modelscope/data-juicer/pull/660/commits/b635d14fd83a900c04f465ef5fabbdb74059e339

data_juicer/__init__.py
```diff
@@ -5,11 +5,12 @@
 import sys
 
 from loguru import logger
-# allow loading truncated images for some too large images.
```
HYLcool, line 8: This comment is necessary for other developers to know the  reason for these two statements:
```
from PIL import ImageFile
ImageFile.LOAD_TRUNCATED_IMAGES = True
```

Maybe keeping it is better.

data_juicer/ops/mapper/audio_add_gaussian_noise_mapper.py
```diff
@@ -11,7 +11,7 @@
 from ..base_op import OPERATORS, Mapper
 from ..op_fusion import LOADED_AUDIOS
 
-audiomentations = LazyLoader('audiomentations', 'audiomentations')
```
HYLcool, line 14: This OP imports `soundfile` explicitly on line 4. Maybe we need to convert it to a lazy-loading way as well.

data_juicer/core/data/ray_dataset.py
```diff
@@ -262,10 +264,10 @@ def read_json(cls, paths: Union[str, List[str]]) -> RayDataset:
             js.open_json
             return read_json_stream(paths)
         except AttributeError:
-            return rd.read_json(paths)
+            return ray.data.read_json(paths)
 
 
-class JSONStreamDatasource(ds.JSONDatasource):
+class JSONStreamDatasource(ray.data.read_api.JSONDatasource):
```
HYLcool, line 270: Here "ray" is used to define this class. Is the lazy-loading of ray always activated by it, even just initializing the DJ from a minimal env?

Maybe need to try to run the basic demo in a minimal env to check it.

data_juicer/ops/deduplicator/ray_basic_deduplicator.py
```diff
@@ -1,18 +1,16 @@
...
 from ..base_op import Filter
 
-redis = LazyLoader('redis', 'redis')
+ray = LazyLoader('ray')
+redis = LazyLoader('redis')
 
 MERSENNE_PRIME = (1 << 61) - 1
 
 
-@ray.remote(scheduling_strategy='SPREAD')
```
HYLcool, line 15: Please @pan-x-c help check if this modification is OK.

- Remove this ray.remote decorator on line 15
- Initialize the RemoteDedupSet explicitly on line 53-56

data_juicer/utils/unittest_utils.py
```diff
@@ -4,15 +4,13 @@
...
-import numpy
-import ray.data as rd
-
 from data_juicer import is_cuda_available
 from data_juicer.core.data import DJDataset, NestedDataset
 from data_juicer.utils.lazy_loader import LazyLoader
 from data_juicer.utils.model_utils import free_models
 
-transformers = LazyLoader('transformers', 'transformers')
+numpy = LazyLoader('numpy')
```
HYLcool, line 12: lazy-loading numpy is not necessary.

data_juicer/ops/deduplicator/ray_basic_deduplicator.py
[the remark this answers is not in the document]
HYLcool: Double checked by @pan-x-c .👌🏻👍🏻

pyproject.toml
```diff
@@ -0,0 +1,193 @@
...
+
+# AI Services & APIs
+ai_services = [
+    "dashscope",  # Alibaba Cloud AI
+    "openai",  # OpenAI API
+    "label-studio==1.17.0",  # Data labeling
+]
+
+# Document Processing
+document = [
```
HYLcool, line 133: This set is already included in the minimal dependencies. Can be erased.

.github/workflows/unit-test-partial.yml
```diff
@@ -81,8 +87,8 @@ jobs:
     - name: Install data-juicer
       working-directory: dj-${{ github.run_id }}/.github/workflows/docker
       run: |
-        docker compose exec ray-head pip install -e .\[all\]
-        docker compose exec ray-worker pip install -e .\[all\]
+        docker compose exec ray-head bash -c '/root/.local/bin/uv venv /workspace/.venv && source /workspace/.venv/bin/activate && /root/.local/bin/uv pip install -e .\[all\]'
```
HYLcool, line 90: The unittest-dist job failed due to `wordcloud` in the minimal requirements was not found (see [log](https://github.com/modelscope/data-juicer/actions/runs/15126362027/job/42519047131?pr=660#step:6:1506)). Please re-check this installing step.

pyproject.toml
```diff
@@ -0,0 +1,193 @@
...
+description = "Data Processing for and with Foundation Models."
+authors = [
+    {name = "SysML Team of Alibaba Tongyi Lab"}
+]
+readme = "README.md"
+license = {text = "Apache-2.0"}
+requires-python = ">=3.10"
+
+# Core dependencies
+dependencies = [
```
HYLcool, line 13: There is a missing dep `scipy`. We can add it to either minimal dependencies or generic set.

HYLcool: Besides, the `cmake>=3.10,<4.0` dep is removed, which is used to limit the cmake version to build `kenlm` and `samplerate` successfully. This issue is solved by their maintainer supposedly. But please double check that this removal is OK when installing them from a clean env.

pyproject.toml
```diff
@@ -0,0 +1,193 @@
...
+    "py-data-juicer[distributed]",
+    "py-data-juicer[dev]",
+    "py-data-juicer[ai_services]",
+    "py-data-juicer[document]",
+]
+
+[project.scripts]
+dj-process = "data_juicer.tools.process_data:main"
+dj-analyze = "data_juicer.tools.analyze_data:main"
+dj-install = "data_juicer.tools.dj_install:main"
```
HYLcool, line 177: The actual locations of these scripts are outside the package "data_juicer". Do we need a similar strategy like "package_dir" in the original setup.py, or does it work fine in this way (After installing, the `dj-process` command is available)?

README.md
```diff
@@ -198,43 +198,144 @@ Table of Contents
...
 
-- Run the following commands to install the latest basic `data_juicer` version in
-  editable mode:
-```shell
-cd <path_to_data_juicer>
-pip install -v -e .
+Data-Juicer is now available on PyPI. The minimal installation includes core data processing capabilities:
+
+```bash
+pip install data-juicer
```
HYLcool, line 214: pip install py-data-juicer

README.md
```diff
@@ -198,43 +198,144 @@ Table of Contents
...
+source .venv/bin/activate  # Unix/macOS
+# or
+.venv\Scripts\activate     # Windows
+
+# Install dependencies
+uv pip install -e ".[desired_components]"
+```
+
+### Scenario-based Installation
+For component details, plz refer to [pyproject.toml](pyproject.toml).
```
HYLcool, line 254: Better to list these optional components in a table like before. It's a little unfriendly to guide users to check in the code.

README.md
```diff
@@ -198,43 +198,144 @@ Table of Contents
...
+
+```bash
+# Distributed Computing
+uv pip install "py-data-juicer[distributed]"
+
+# AI Services & APIs
+uv pip install "py-data-juicer[ai_services]"
+
+# Utilities (Web, API, Visualization)
+uv pip install "py-data-juicer[utils]"
```
HYLcool, line 289: No "utils" anymore.

README_ZH.md
```diff
@@ -189,109 +189,95 @@ Data-Juicer正在积极更新和维护中，我们将定期强化和新增更多
...
-
-### 从源码安装 (指定使用场景)
-
-* 运行以下命令以安装 `data_juicer` 可编辑模式的最新基础版本
-
-```shell
-cd <path_to_data_juicer>
-pip install -v -e .
+# 或使用 uv 进行更快的安装
+uv pip install data-juicer
```
HYLcool, line 199: On lines 196 and 199, "data-juicer" --> "py-data-juicer"

README.md
```diff
@@ -198,43 +198,144 @@ Table of Contents
...
+```
+
+
+**5. For Development Mode**
+
+For contributors and developers:
+
+```bash
+# Clone repository
+git clone https://github.com/alibaba/data-juicer.git
```
HYLcool, line 329: https://github.com/alibaba/data-juicer.git
-->
https://github.com/modelscope/data-juicer.git

README_ZH.md
```diff
@@ -189,109 +189,95 @@ Data-Juicer正在积极更新和维护中，我们将定期强化和新增更多
...
-```shell
-cd <path_to_data_juicer>
-pip install -v -e .  # 安装最小依赖，支持基础功能
-pip install -v -e .[tools] # 安装部分工具库的依赖
-```
-
-依赖选项如下表所示:
+```bash
+# 克隆仓库
+git clone https://github.com/alibaba/data-juicer.git
```
HYLcool, line 206: https://github.com/alibaba/data-juicer.git
-->
https://github.com/modelscope/data-juicer.git

README_ZH.md
```diff
@@ -189,109 +189,95 @@ Data-Juicer正在积极更新和维护中，我们将定期强化和新增更多
...
-| 标签              | 描述 |
-|------------------|----------------------------------|
-| `.` 或 `.[mini]` | 为基本 Data-Juicer 安装最小依赖项。  |
-| `.[all]`         | 为除沙盒之外的所有 OP 安装依赖项。    |
-| `.[sci]`         | 为与科学用途相关的 OP 安装依赖项。    |
-| `.[dist]`        | 安装用于分布式数据处理的额外依赖项。   |
-| `.[dev]`         | 安装作为贡献者开发软件包的依赖项。     |
-| `.[tools]`       | 安装专用工具（例如质量分类器）的依赖项。|
-| `.[sandbox]`     | 安装沙盒的所有依赖项。               |
+# 使用 pip 安装所有依赖
```
HYLcool, line 209: This Chinese version of this part should be aligned with the English version.

README_ZH.md
```diff
@@ -189,109 +189,95 @@ Data-Juicer正在积极更新和维护中，我们将定期强化和新增更多
...
-随后在终端运行 ffmpeg 命令检查环境是否设置正确。
+# 导出依赖列表
+uv pip freeze > requirements.txt
 
+# 从 requirements.txt 安装依赖
+uv pip install -r requirements.txt
+```
 
-<p align="right"><a href="#table">🔼 back to index</a></p>
+使用 `uv` 进行虚拟环境管理的优势：
```
HYLcool, line 276: No need to highlight the advatages for 3rd-party lib.

tools/dj_install.py
```diff
@@ -1,64 +1,45 @@
...
-            f'An error occurred while installing the requirements: {e}')
-        if os.path.exists(temp_file_path):
-            os.remove(temp_file_path)
-        sys.exit(1)
-    finally:
-        if os.path.exists(temp_file_path):
-            os.remove(temp_file_path)
+    parser = argparse.ArgumentParser(
+        description='Install Data-Juicer dependencies')
+    parser.add_argument(
```
HYLcool, line 12: I think it's better to keep both the original installation way (config-wise) and this new way (category-wise).

The motivation of the previous way is that users often don't know which categories are required by their config recipes, so we provide a convenient way to install dependencies according to OPs in the config files.

docs/DeveloperGuide.md
```diff
@@ -393,13 +393,110 @@ class PerplexityFilter(Filter):
...
+
+# Optional: lock up dependency versions
+# Export dependencies to requirements.txt
+uv pip freeze > requirements.txt
+
+# Install dependencies from requirements.txt
+uv pip install -r requirements.txt
+```
+
+Advantages of using `uv` for virtual environment management:
```
HYLcool, line 449: No need to list these advantages for 3rd-party libs.

docs/DeveloperGuide.md
```diff
@@ -393,13 +393,110 @@ class PerplexityFilter(Filter):
...
+2. The lazy loading system will automatically handle installation when the dependencies are first used.
+
+Example:
+```toml
+[project.dependencies]
+# Core dependencies
+numpy = ">=1.26.4,<2.0.0"
+
+[project.optional-dependencies]
+sci = [
```
HYLcool, line 472: Change the sci category to a updated one.

docs/DeveloperGuide.md
```diff
@@ -393,13 +393,110 @@ class PerplexityFilter(Filter):
...
+### Development Setup
+
+1. Install the package with all dependencies:
+```bash
+uv pip install -e ".[all]"
+```
+
+2. Or install with specific groups:
+```bash
+uv pip install -e ".[sci]"    # Scientific dependencies
```
HYLcool, line 487: Need to align with our latest groups.

docs/DeveloperGuide_ZH.md
```diff
@@ -463,3 +464,99 @@ optional arguments:
 ......
 
 ```
+
+## 依赖管理
```
HYLcool, line 468: Refer to the comments in the English version.

data_juicer/utils/lazy_loader.py
```diff
@@ -1,69 +1,437 @@
...
-                                                  _is_module_installed)
-from data_juicer.utils.availability_utils import _torch_check_and_set
 
-current_path = os.path.dirname(os.path.realpath(__file__))
-science_file_path = os.path.join(current_path,
-                                 '../../environments/science_requires.txt')
-dist_file_path = os.path.join(current_path,
-                              '../../environments/dist_requires.txt')
-AUTOINSTALL = AutoInstaller([science_file_path, dist_file_path])
+def get_toml_file_path():
```
HYLcool, line 15: This might not work when installing through 'pypi', because "pyproject.toml" usually won't be packed in the pypi file. (same issue in previous way).

data_juicer/utils/lazy_loader.py
```diff
@@ -1,69 +1,437 @@
...
+        if isinstance(pip_args, str):
+            pip_args = [pip_args]
+
+        # For GitHub repositories, clone only to get dependencies
+        if package_spec.startswith(('git+', 'https://github.com/')):
+            import os
+            import shutil
+            import tempfile
+
+            import git
```
HYLcool, line 216: Do we need to add `git` to dependency list?

.github/workflows/unit-test-partial.yml
cyruszhang: it's because when running code in ray-worker venv is not by default activated.  instead of littering the github workflows with venv, we will stick with system python

data_juicer/core/data/ray_dataset.py
cyruszhang: confirmed with a clean minimal install, test_process_data.py works well; ray is properly lazy-loaded

pyproject.toml
cyruszhang: > Besides, the `cmake>=3.10,<4.0` dep is removed, which is used to limit the cmake version to build `kenlm` and `samplerate` successfully. This issue is solved by their maintainer supposedly. But please double check that this removal is OK when installing them from a clean env.

Confirmed that kenlm and samplerate are all working now. Also, upon inspection, samplerate is preferred since its efficient C implementation is suitable for large scale processing - resampy is much slower even in local test.

pyproject.toml
cyruszhang: hatchling handles this logic.

data_juicer/utils/lazy_loader.py
cyruszhang: gitpython is installed via other dependencies. but I will call it out in minimal dep

data_juicer/utils/lazy_loader.py
cyruszhang: good insight. will test

cyruszhang: resolved in https://github.com/modelscope/data-juicer/pull/660/commits/fa47f5e7b651264c9f6ea5b822f7a6596743c035

- use uv.lock for deterministic dependency resolution
- distribute uv.lock with wheels

tools/dj_install.py
cyruszhang: resolved in this https://github.com/modelscope/data-juicer/pull/660/commits/ef4ab3a2ff8577a1e48641206aedb1e94d028f14:
- use ast to check op imports
- use LazyLoader.check_packages to handle necessary installations 
- bring back centralized module to package mapping to rid redundancy when using LazyLoader and deal with dj_install dependency resolution

yxdyc (approved):
LGTM

HYLcool:
Some workflows are broken due to the usage of `uv` and `bash -c`:
- performance benchmark: https://github.com/modelscope/data-juicer/actions/runs/15176858936/job/42678639134
- doc building: https://github.com/modelscope/data-juicer/actions/runs/15176858925/job/42678639205

yxdyc:
> Some workflows are broken due to the usage of `uv` and `bash -c`:
> 
> * performance benchmark: https://github.com/modelscope/data-juicer/actions/runs/15176858936/job/42678639134
> * doc building: https://github.com/modelscope/data-juicer/actions/runs/15176858925/job/42678639205
> * publish pypi package might fail as well.

update: being resolved by @cmgzn

PR #663, opened 2025-05-07
Add correlation_analysis in the analysis module

HYLcool:
As the title says.


yxdyc:
It's suggested that we mention this feature somewhere (e.g., in our analyzer by default, the sandbox, or in some documentation). Otherwise, it will remain an 'orphan node' in the repo.

Others LGTM.

HYLcool:
> It's suggested that we mention this feature somewhere (e.g., in our analyzer by default, the sandbox, or in some documentation). Otherwise, it will remain an 'orphan node' in the repo.
> 
> Others LGTM.

Already added it to the default pipeline of Analyzer. Besides, described what information will be produced by Analyzer in the QuickStart docs.

PR #666, opened 2025-05-09
Split standalone and dist unit tests into different jobs

HYLcool:
As the title says.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -35,7 +35,6 @@ jobs:
       working-directory: dj-${{ github.run_id }}/.github/workflows/docker
       run: |
         docker compose exec ray-head pip install -e .\[all\]
-        docker compose exec ray-worker pip install -e .\[all\]
 
     - name: Clean dataset cache
       working-directory: dj-${{ github.run_id }}/.github/workflows/docker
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-test.yml
```diff
@@ -60,6 +59,41 @@ jobs:
...
+      if: always()
+      run: |
+        docker compose down --remove-orphans
+
+    - name: Cleanup workspace
+      if: always()
+      run: |
+        rm -rf  dj-${{ github.run_id }}
+
+  unittest-dist:
```
yxdyc, line 73: In the coverage part, https://github.com/modelscope/data-juicer/blob/f2bf393fe61139e78c6138ace83454152fa8d523/.github/workflows/unit-test.yml#L122, the result is based on previous `unittest-single`. After the splitting, do we need to modify the logic to incorporate `unittest-dist` accordingly?

HYLcool: I think it's ok. The coverage report mainly focuses on the basic functional coverage, which comes from standalone unit tests.

If It's necessary, we can report two coverage results for both basic functionality and distributed one in the future.

yxdyc (approved):

PR #668, opened 2025-05-09
Ignore the installed distutils libs for docker image building

HYLcool:
As the title says.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -55,6 +55,12 @@ jobs:
         with:
           images: ${{ env.IMAGE_NAME }}
 
+      - name: Get the current version tag
+        id: get-version-tag
+        working-directory: dj-${{ github.run_id }}
+        run: |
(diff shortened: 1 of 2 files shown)


cyruszhang:
Seems --ignore-installed will ignore already installed versions, skip uninstalling, and overwrite with newer versions; could cause orphaned obsolete versions. What triggered us to adopt this?

HYLcool:
> Seems --ignore-installed will ignore already installed versions, skip uninstalling, and overwrite with newer versions; could cause orphaned obsolete versions. What triggered us to adopt this?

There is a dependency "blinker" that is first installed by `apt install` with an older version (1.4) and then required by the latest streamlit for a newer version (>1.5). But uninstalling the older version is not allowed and raises an error "It is a distutils installed project and thus we cannot accurately determine which files belong to it, which would lead to only a partial uninstall."

I found two ways to resolve this issue:
1. Install the newer version with "--ignore-installed"
2. Limit the streamlit version to "<1.45.0"

Here I just choose the 1st way to avoid limiting the version of some other more important dependencies.

cyruszhang:
LGTM

cyruszhang (approved):
LGTM

PR #679, opened 2025-05-22
adapt the resources count of ray and fix calculate_np and fix mem_required set

Cathy0908:


data_juicer/utils/process_utils.py
```diff
@@ -49,6 +51,124 @@ def get_min_cuda_memory():
...
+        available_mems = []
+        for nodeid, info in ray_nodes_info:
+            available_mems.append(info['memory'])
+
+        return available_mems
+
+    return [psutil.virtual_memory().available]
+
+
+def available_gpu_memories():
```
cyruszhang: besides memories, is number of GPUs logic also being updated?

Cathy0908: the returned is the available memory of each gpu card for each alive node, is there no need to add the number of gpus ?

HYLcool:
The OOM issue in the unittest-dist will be addressed in the future PRs.

PR #681, opened 2025-05-23
Add support for text embedding via API and local model

lingzhq:
1. Add **prepare_embedding_model()** for local model to perform text embedding tasks. Usage: model.encode(text) returns an embedding list.
2. Modify the existing **prepare_api_model()**. The original API only supported chat calls; this update enables API calls based on endpoint recognition for two types (chat, embedding) and leaves room for potential future API call methods.
3. Renamed the original APIModel class to **ChatAPIModel class**. The usage logic remains unchanged, naturally supporting existing call methods, and **filter_arguments()** is refactored as a standalone function outside the class.
4. Introduced **EmbeddingAPIModel class** for API calls related to text embedding. Usage: model(input, ...) returns an embedding list.

### DIFF
--- a/data_juicer/utils/model_utils.py
+++ b/data_juicer/utils/model_utils.py
@@ -127,7 +127,27 @@ def check_model(model_name, force=False):
     return cached_model_path
 
 
-class APIModel:
+def filter_arguments(func, args_dict):
+    """
+    Filters and returns only the valid arguments for a given function
+    signature.
+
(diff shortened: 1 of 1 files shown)


HYLcool (approved):
LGTM. Please add corresponding unit tests in tests/utils/test_model_utils.py if possible. And please @yxdyc double-check this PR.

PR #684, opened 2025-05-27
optimize startup time

cyruszhang:
optimize dj-process startup time from ~80seconds to ~6 seconds (10X speed up) by:

1. argparse.parse_args uses two-phase argument parsing:
- Essential arguments are handled as usual.
- OP optional arguments are only imported when the corresponding OP is used.

2. update_op_process:
- Only processes arguments for the OPs that are actually used.
- Several optimizations: replaced list comprehensions with sets, removed unnecessary deep copies, and eliminated redundant type checks.

confirmed the generated config table are exactly the same.

### DIFF
--- a/data_juicer/config/config.py
+++ b/data_juicer/config/config.py
@@ -6,6 +6,7 @@ import shutil
 import tempfile
 import time
 from argparse import ArgumentError
+from contextlib import contextmanager
 from typing import Dict, List, Optional, Union
 
 import yaml
@@ -24,6 +25,14 @@ global_cfg = None
 global_parser = None
 
 
+@contextmanager
+def timing_context(description):
+    start_time = time.time()
+    yield
+    elapsed_time = time.time() - start_time
+    logger.info(f'{description} took {elapsed_time:.2f} seconds')
+
+
 def init_configs(args: Optional[List[str]] = None, which_entry: object = None):
     """
     initialize the jsonargparse parser and parse configs from one of:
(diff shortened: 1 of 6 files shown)


data_juicer/config/config.py
```diff
@@ -644,113 +730,98 @@ def sort_op_by_types_and_names(op_name_classes):
...
+    # Only include recognized args
+    for key, value in vars(temp_cfg).items():
+        if key in recognized_args and value is not None:
+            temp_args.extend([f'--{key}', str(value)])
 
-    for key, value in vars(namespace).items():
-
-        if issubclass(type(value), Namespace):
-            nested_args = namespace_to_arg_list(value, f'{prefix}{key}.')
-            arg_list.extend(nested_args)
```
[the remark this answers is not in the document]
cyruszhang: missing test cases for regression

data_juicer/config/config.py
```diff
@@ -482,6 +482,8 @@ def init_configs(args: Optional[List[str]] = None, which_entry: object = None):
             with timing_context('Parsing arguments'):
                 cfg = parser.parse_args(args=args)
 
+                logger.info(f'cfg: {cfg}')
```
HYLcool: This redundant log can be removed

data_juicer/ops/base_op.py
```diff
@@ -319,15 +319,11 @@ def __init__(self, *args, **kwargs):
                 skip_op_error=self.skip_op_error,
                 op_name=self._name)
 
-    # set the process method is not allowed to be overridden
+    @classmethod
     def __init_subclass__(cls, **kwargs):
-        not_allowed_list = ['process']
```
yxdyc, line 324: Maintaining `not_allowed_list = ['process']` maybe necessary, for standard OP customization?

data_juicer/ops/base_op.py
```diff
@@ -319,15 +319,11 @@ def __init__(self, *args, **kwargs):
...
-        for method_name in not_allowed_list:
-            if method_name in cls.__dict__:
-                raise TypeError(
-                    f'Method {method_name} cannot be overridden by subclass '
-                    f'{cls.__name__}. Please implement {method_name}_single '
-                    f'or {method_name}_batched.')
+        """Register the operator to the registry."""
+        super().__init_subclass__(**kwargs)
+        OPERATORS.register_module(cls.__name__.lower())(cls)
```
yxdyc: Difference between this implementation and adding wrapper before class definition?   Does this new modification introduce double registration?

data_juicer/ops/base_op.py
```diff
@@ -401,6 +397,7 @@ def __init__(self, *args, **kwargs):
         """
         super(Filter, self).__init__(*args, **kwargs)
         self.stats_export_path = kwargs.get('stats_export_path', None)
+        self._stats = {}
```
yxdyc: Some Filters do not produce `_stats`, e.g., `specified_field_filter` & `specified_field_filter `, which depend on `_stats` produced by other OPs.

data_juicer/ops/base_op.py
```diff
@@ -423,15 +420,12 @@ def __init__(self, *args, **kwargs):
...
-        for method_name in not_allowed_list:
-            if method_name in cls.__dict__:
-                raise TypeError(
-                    f'Method {method_name} cannot be overridden by subclass '
-                    f'{cls.__name__}. Please implement {method_name}_single '
-                    f'or {method_name}_batched.')
+        """Register the filter to the registry."""
+        super().__init_subclass__(**kwargs)
+        if not hasattr(cls, '_stats'):
+            NON_STATS_FILTERS.register_module(cls.__name__.lower())(cls)
```
yxdyc: The line 400 makes `NON_STATS_FILTERS` invalid (always empty)

yxdyc (commented):
I have some questions about the `base_op.py`, plz see the inline comments. If the descriptions are not clear, we can clarify more details (e.g., with yilun, or in DingTalk).

data_juicer/ops/base_op.py
cyruszhang: good point. missed logic

HYLcool (approved):
LGTM

PR #694, opened 2025-06-09
MinHash calculation with GPU on Ray

cyruszhang:
- use GPU for MinHash calculation (accelerator: 'cuda')
- automatic resolution of concurrency and batch size
- ray CPU/GPU and status checking utility 
- add demo configs
- update dependency


data_juicer/utils/resource_utils.py
```diff
@@ -50,3 +51,66 @@ def query_mem_info(query_key):
...
+        else:
+            logger.info(f'Found {gpu_count} GPUs in Ray cluster')
+
+        return gpu_count
+    except Exception as e:
+        logger.error(f'Error getting Ray GPU count: {str(e)}')
+        return 0
+
+
+def get_ray_gpu_memory():
```
cyruszhang, line 83: currently in calculate_np we only consider local GPU memory; this utility gives up more GPU insights and enable more automatic resource resolution on ray cluster.

data_juicer/ops/deduplicator/ray_bts_minhash_deduplicator.py
```diff
@@ -566,22 +671,77 @@ def minhash_with_uid(table: pa.Table) -> pa.Table:
...
+                raise RuntimeError('No GPUs available in Ray cluster')
+
+            concurrency = max(1,
+                              gpu_count)  # Ensure at least 1 concurrent task
+            logger.info(
+                f'Setting GPU concurrency to {concurrency} based on available GPUs'
+            )
+
+            # Get available GPU memory and set batch size
+            gpu_memory = get_ray_gpu_memory()
```
cyruszhang, line 689: should be able to use https://github.com/modelscope/data-juicer/pull/679/files if merged

.gitignore
```diff
@@ -15,6 +15,7 @@ dist
 wandb/
 __pycache__
 .vscode/
+.ipynb_checkpoints/
```
HYLcool, line 18: Already added in the main branch

pyproject.toml
```diff
@@ -81,6 +81,7 @@ vision = [
     "scenedetect[opencv]",  # Scene detection
     "ultralytics",  # YOLO models
     "rembg",  # Background removal
+    "cudf-cu12",
```
HYLcool: Maybe in the "distributed" set?

cyruszhang: cudf is not inherently distributed; it can be used on single GPU. in other OPs we should be able to utilize in a non-ray scenario

cyruszhang: dask is the distributed cudf

pyproject.toml
```diff
@@ -122,13 +122,7 @@ dev = [
     "pre-commit",  # pre-commit hooks
     "sphinx",
     "sphinx-autobuild",
-    "sphinx-multiversion",
```
HYLcool: These 6 deps are necessary for doc building.

pyproject.toml
```diff
@@ -154,7 +148,20 @@ sandbox = [
     "torch>=1.11.0",
     "fire",
     "pyspark",
-    "modelscope[framework,nlp]",
+    "vbench",
```
HYLcool: And these 14 deps are removed already from the sandbox group

HYLcool: Looks like a broken merge from the main branch.

cyruszhang: indeed. fixed

PR #687, opened 2025-06-04
support RayExporter

Cathy0908:

PR #697, opened 2025-06-11
Add file extension detection when using ray executor.

chenyushuo:
Implement automatic file extension detection for Ray executor
﻿
This PR introduces a new feature for the Ray executor that automatically detects file extensions when processing input data. Here are the key points:
﻿
1. Automatic Format Detection:
- If `dataset.configs.source` is not explicitly set, the Data-Juicer will now automatically recognize the file extension and use the corresponding data format.
﻿
2. Manual Format Override:
- Users still have the option to force a specific data format by setting the `dataset.configs.source` parameter.
﻿
3. Enhanced Flexibility:
- This change improves the handling of various data sources when using the Ray executor.
- It supports multiple data formats, including but not limited to JSON and Parquet.
﻿
4. User Benefits:
- Reduces the need for manual configuration in many cases.
- Provides a more intuitive and user-friendly experience when working with different file types.
﻿

This enhancement aims to streamline data processing workflows and accommodate a wider range of input formats with minimal user intervention.
﻿


data_juicer/core/data/load_strategy.py
```diff
@@ -239,13 +239,63 @@ def load_data(self, **kwargs):
...
+            '.jsonl': 'json',
+            '.txt': 'text',
+            '.csv': 'csv',
+            '.tsv': 'csv',
+            '.parquet': 'parquet',
+            '.npy': 'numpy',
+            '.tfrecords': 'tfrecords',
+        }
+        auto_detect = False
+        data_source = self.ds_config.get('source', None)
```
cyruszhang, line 255: maybe make executor_type /  data_type /  data_source or strategy_key part of DataloadStrategy base class? that way we can use self.data_source of self.key.data_source

data_juicer/core/data/load_strategy.py
```diff
@@ -239,13 +239,63 @@ def load_data(self, **kwargs):
                     f'Current working directory: {os.getcwd()}')
 
         logger.info(f'Using resolved path for loading ray dataset: {path}')
+
+        file_extension_map = {
+            '.json': 'json',
```
cyruszhang, line 244: add lance since it's in demand https://github.com/modelscope/data-juicer/pull/687

data_juicer/core/data/load_strategy.py
```diff
@@ -239,13 +239,63 @@ def load_data(self, **kwargs):
...
+        else:
+            suffix = os.path.splitext(data_source)[1]
+            if suffix in file_extension_map:
+                data_format = file_extension_map[suffix]
+            elif '.' + data_source in file_extension_map:
+                data_format = file_extension_map['.' + data_source]
+            else:
+                auto_detect = True
+        if auto_detect:
+            item_path = path
```
cyruszhang, line 267: first seen file defines the data reader; make sure we manifest that in the comment

data_juicer/core/data/load_strategy.py
chenyushuo: OK, added

data_juicer/core/data/load_strategy.py
chenyushuo: OK, added

cyruszhang:
LGTM

PR #709, opened 2025-06-18
[NewOp] Add download_file_mapper op

Cathy0908:


data_juicer/ops/mapper/download_file_mapper.py
```diff
@@ -0,0 +1,102 @@
...
+        self.retry_delay = retry_delay
+        self.max_delay = max_delay
+        self.stream = stream
+        self.headers = headers
+
+    def _download_data_with_context(self, sample, context):
+        """
+        Download files with contexts.
+        """
+        raw_urls = sample[self.download_field]
```
cyruszhang: validate field presence? add test case as well

cyruszhang: could be a list or a single url? handle both

data_juicer/ops/mapper/download_file_mapper.py
```diff
@@ -0,0 +1,102 @@
...
+        new_paths = []
+        response = None
+
+        for raw_url in raw_urls:
+            if is_remote_path(raw_url):
+                save_path = osp.join(self.save_dir, osp.basename(raw_url))
+                if not osp.exists(save_path):
+                    response = download_file(raw_url,
```
cyruszhang: should we use coroutine or some sort of async process for a blocking action like file download?

cyruszhang: daft uses async for downloading actions as well: https://raw.githubusercontent.com/Eventual-Inc/Daft/e984a22e1a69c78e0745cd30fdc3fd3ff6e29b37/src/daft-functions-uri/src/download.rs

data_juicer/ops/mapper/download_file_mapper.py
Cathy0908: done, field check in self.process_single

data_juicer/ops/mapper/download_file_mapper.py
Cathy0908: done

data_juicer/ops/mapper/download_file_mapper.py
```diff
@@ -0,0 +1,146 @@
...
+
+        async def run_downloads(urls, save_dir, **kwargs):
+            async with aiohttp.ClientSession() as session:
+                tasks = [
+                    _download_file(session, idx, url, save_dir, **kwargs)
+                    for idx, url in enumerate(urls)
+                ]
+                return await asyncio.gather(*tasks)
+
+        results = asyncio.run(run_downloads(urls, save_dir, **kwargs))
```
cyruszhang, line 84: good stuff.  should we limit concurrency when downloading? using asyncio semaphore

Cathy0908: done

cyruszhang (approved):
LGTM

PR #710, opened 2025-06-19
Fix sandbox left bugs

HYLcool:
- Add new hook for converting dataset format for easyanimate
- Make easyanimate pipelines in the sandbox work
- Fix bugs in log summarization


yxdyc (approved):
LGTM

PR #717, opened 2025-06-24
Fix bugs found in regression tests

HYLcool:
- fix spacy model loading bugs.
- downgrade torch/vllm to make them compatible with each other
- make a stronger free_models func to avoid OOM during unit test
- add `torch.no_grad` for some video OPs.
- others:
  - fix a bad link in the docs
  - print the dependency tree after installing the environment for checking conveniently.

- Successful regression test: https://github.com/modelscope/data-juicer/actions/runs/15845628668


cyruszhang:
run tools/generate_uv_lock.py to update package versions. do we still need to get rid of sandbox dependencies when pip install .[all]?

data_juicer/ops/mapper/video_captioning_from_frames_mapper.py
```diff
@@ -232,15 +232,16 @@ def _process_single_sample(self, ori_sample, rank=None, context=False):
...
-                    for i in range(self.caption_num):
-                        generated_ids = model.generate(**inputs,
-                                                       max_new_tokens=128,
-                                                       do_sample=True)
-                        generated_text = processor.batch_decode(
-                            generated_ids, skip_special_tokens=True)
-                        generated_text_candidates_single_chunk[i] += [
-                            '. '.join([txt.strip() for txt in generated_text])
-                        ]
+                    with torch.no_grad():
```
cyruszhang, line 235: I assume for inference, no grad calculation is by default; does this line change the behavior of the tests?

HYLcool: I tried the version without `torch.no_grad` and the requires_grad of output logits is True. So I think it's not on by default. This line is not the core solution to avoid OOM error during unit test, but it helps to reduce the cuda memory as well. The new version of free_models func is the core solution.

HYLcool:
> run tools/generate_uv_lock.py to update package versions. do we still need to get rid of sandbox dependencies when pip install .[all]?

`sandbox` dependencies are added to `all` now, and a new `uv.lock` file is generated. Perhaps we can add the generation of uv.lock to the pre-commit process in the future.

cyruszhang (approved):
LGTM

PR #721, opened 2025-06-26
[WIP] add lidar object detection op

Cathy0908:


data_juicer/utils/model_utils.py
```diff
@@ -942,6 +943,57 @@ def update_sampling_params(sampling_params, pretrained_model_name_or_path, enabl
     return sampling_params
 
 
+class MMLabModel(object):
```
yxdyc, line 946: FYI, @Qirui-jiao : based on onnx models from mmdeploy

PR #724, opened 2025-07-01
update DownloadFileMapper and custom webdataset encoder and decoder

Cathy0908:
…and decoder


tests/core/test_ray_exporter.py
```diff
@@ -108,6 +109,54 @@ def test_lance_keep_hashes(self):
 
         self.assertListEqual(data_list, self._pop_raw_data_keys([Fields.stats]))
 
+    def test_webdataset_multi_images(self):
```
HYLcool, line 112: Decorate it with `@TEST_TAG('ray')`

PR #725, opened 2025-07-02
Support store image bytes in the dataset

HYLcool:
- Only for images for now.
- Add a new field "image_bytes" to store bytes data of images.
- When loading the PIL.Image objects, load them from bytes first if they exist. (in load_data_with_context func)
- Update the invoking way of load_data_with_context for ops that use this func. For some OPs that might generate new image contents, update the bytes data according to the new contents.
- Support context sharing in the GeneralFusedOP


HYLcool:
Regression tests on this PR are passed.
https://github.com/modelscope/data-juicer/actions/runs/16018627756

HYLcool:
Merge PR #724 into this branch.

HYLcool:
- Support customized format before exporting for webdataset
- Support export_type arg to specify the export format type instead of parsing from the export_path

data_juicer/format/formatter.py
```diff
@@ -212,9 +211,9 @@ def non_empty_text(sample, target_keys):
                 ds_dir = os.path.dirname(global_cfg.dataset_path)
         else:
             ds_dir = ""
-        image_key = global_cfg.get("image_key", SpecialTokens.image)
-        audio_key = global_cfg.get("audio_key", SpecialTokens.audio)
-        video_key = global_cfg.get("video_key", SpecialTokens.video)
+        image_key = global_cfg.get("image_key", "images")
```
cyruszhang: why the reversion?

data_juicer/core/ray_exporter.py
```diff
@@ -90,6 +93,16 @@ def _export_impl(self, dataset, export_path, columns=None):
 
         if self.export_format in {"json", "jsonl"}:
```
cyruszhang: should we adopt use _router() mechanics like in exporter? 
why is webdataset not supported in local mode?

cyruszhang (commented):
what is the over-arching use case for this change/feature?

data_juicer/format/formatter.py
HYLcool: Here is to set the field key name to store these multimodal data instead of the special tokens in the texts.

data_juicer/core/ray_exporter.py
HYLcool: _router() mechanics done.

HF datasets don't support conversion to webdataset format for now.

PR #730, opened 2025-07-04
[bug fix] Fix bug for issue #711: the parameters will be overwritten by the same operator

ricksun2023:
If we put the same operator many times in a _process.yaml_, the parameters of the last operator can be used to executed, but the previous ones will be overwritten by the last one.
So, we need a list to save all parameters of all entities of the same operator.

### DIFF
--- a/data_juicer/config/config.py
+++ b/data_juicer/config/config.py
@@ -787,21 +787,27 @@ def update_op_process(cfg, parser, used_ops=None):
     op_name_count = {}
     for op_name in used_ops:
         op_config = op_configs.get(op_name)
-
+        op_config_list = []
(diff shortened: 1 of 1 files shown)


ricksun2023:
@yxdyc @HYLcool @cyruszhang Would you review this code for #711 bug fix？

HYLcool:
Hi @ricksun2023 , you might need to run the corresponding unit tests locally and check if the modified code works or if you need to update the unit test cases:
```shell
python tests/config/test_config.py
```

HYLcool (approved):
LGTM. Thanks for your contribution.👍🏻

PR #739, opened 2025-07-17
Support shard_size and extra args for write methods in  export_extra_args for RayExporter

HYLcool:
As the title says.


yxdyc (approved):
LGTM

PR #761, opened 2025-08-19
Add Operator-Level Parallel Data Processing with Ray Actors

Cccccc0630:
This PR introduces significant performance optimizations for parallel data processing using Ray Actors and multi-threading. The following changes have been implemented:

1. **Dynamic Resource Allocation**: 
   - Multiple Actors per operator (OP) are now dynamically created based on resource requirements (CPU/GPU).
   - The OP using CUDA will load corresponding model onto to available resources accordingly when its Actors is created.

3. **Parallel Data Processing**: 
   - A data distribution thread is responsible for distributing batches of data to the first operator's actors, supporting data processing for multiple streams.
  
## Benefits:
- **Improved Performance**: By utilizing multi-threading and Ray Actors for parallel data processing, the system can handle large volumes of data more efficiently.
- **Scalability**: The dynamic creation of actors based on resource availability allows the system to scale according to the workload.

## Future Work (Potential Follow-up PRs)​:
- Support parallel batch processing and implement processing_batched for OPs in the pr_demo.yaml process.
- Achieve multi-Actor parallelism on GPUs to improve GPU utilization and SM% (Streaming Multiprocessor percentage).

## Experiment:
The end-to-end efficiency optimization for a small amount of videos is shown in the figure below. Increasing the amount of data can make the advantages of parallel processing more obvious.
juicer drawio2

### DIFF
--- a/data_juicer/core/data/dj_dataset.py
+++ b/data_juicer/core/data/dj_dataset.py
@@ -37,6 +37,10 @@ class DJDataset(ABC):
         """process a list of operators on the dataset."""
 
     @abstractmethod
+    def process_parallel(self, operators, *, exporter=None, checkpointer=None, tracer=None) -> DJDataset: 
+        """Implementing op parallel data processing based on Ray Actor"""
+
+    @abstractmethod
     def schema(self) -> Schema:
         """Get dataset schema.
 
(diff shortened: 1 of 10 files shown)


pan-x-c:
Good job!
Here are some suggestions:
1. Clean up the code, use English comments and fix pre-commit.
2. Move `core/ray_actor.py` to `ops`. The current `Actor` is actually more like an operator wrapper, and can even be renamed to `RayOperatorWrapper` or something like it.

PR #774, opened 2025-09-05
support specifying num_proc, cpu_required and gpu_required for each op

Cathy0908:


data_juicer/core/data/ray_dataset.py
```diff
@@ -176,13 +183,20 @@ def process_batch_arrow(table: pyarrow.Table):
...
                 else:
                     self.data = self.data.map_batches(
-                        op.process, batch_size=batch_size, batch_format="pyarrow", num_gpus=num_gpus
+                        op.process,
+                        batch_size=batch_size,
+                        batch_format="pyarrow",
+                        num_cpus=op.cpu_required,
+                        concurrency=(
+                            None if auto_parallel else op_proc
+                        ),  # use ray default parallelism in cpu mode if num_proc is not specified
```
HYLcool, line 199: num_gpus is missing

data_juicer/core/data/ray_dataset.py
```diff
@@ -203,30 +219,39 @@ def process_batch_arrow(table: pyarrow.Table):
...
                 else:
                     self.data = self.data.map_batches(
-                        op.compute_stats, batch_size=batch_size, batch_format="pyarrow", num_gpus=num_gpus
+                        op.compute_stats,
+                        batch_size=batch_size,
+                        batch_format="pyarrow",
+                        num_cpus=op.cpu_required,
+                        concurrency=(
+                            None if auto_parallel else op_proc
+                        ),  # use ray default parallelism in cpu mode if num_proc is not specified
```
HYLcool, line 235: num_gpus is missing

data_juicer/core/data/ray_dataset.py
Cathy0908: It is under the cpu branch, so gpu is not specified

data_juicer/core/data/ray_dataset.py
Cathy0908: same as above

PR #789, opened 2025-09-19
Optimize the auto num_proc calculation of operators in ray mode

Cathy0908:
参考文档：https://aliyuque.antfin.com/pai-innovative-algo/reep8g/czmqm2vxa0u7arkz


tests/core/data/test_ray_auto_proc.py
cyruszhang, line 1: aoto -> auto

data_juicer/core/data/ray_dataset.py
```diff
@@ -90,6 +91,73 @@ def filter_batch(batch, filter_func):
     return batch.filter(mask)
 
 
+def find_optimal_concurrency(resource_ratios, total_resource):
```
cyruszhang: can we have perf test results to go with the optimization?

Cathy0908: We will conduct detailed perf testing and provide reports afterward.

Cathy0908: reference: https://aliyuque.antfin.com/pai-innovative-algo/reep8g/czmqm2vxa0u7arkz

HYLcool:
The `runtime_np` method in `base_op.py` need to be updated as well. The current unittest-dist is stuck in the test cases of `test_alphanumeric_filter` and I think the reason might be that it doesn't call the right `calculate_ray_np` method in the `runtime_np` method in `base_op.py`. See the [unittest run](https://github.com/modelscope/data-juicer/actions/runs/18368174021/job/52325063591?pr=789#step:6:14035).

image

Cathy0908:
> The `runtime_np` method in `base_op.py` need to be updated as well. The current unittest-dist is stuck in the test cases of `test_alphanumeric_filter` and I think the reason might be that it doesn't call the right `calculate_ray_np` method in the `runtime_np` method in `base_op.py`. See the [unittest run](https://github.com/modelscope/data-juicer/actions/runs/18368174021/job/52325063591?pr=789#step:6:14035).
> 
> image

@HYLcool  The `runtime_np` in the `base_op.py` file only takes effect on a single node. In ray mode, the `run` method of base_op is not used.

HYLcool (approved):

PR #801, opened 2025-10-29
Add video_object_segmenting_mapper and video_depth_estimation_mapper

Qirui-jiao:
Add two video analysis ops, including:

- `video_object_segmenting_mapper`: Performs text-guided semantic segmentation of valid objects throughout the video (using YOLOE and SAM2), with support for saving segmentation visualization results.

- `video_depth_estimation_mapper`: Performs depth estimation on the video, with support for saving both visualization results and point cloud data.

### DIFF
--- a/tests/ops/mapper/test_video_depth_estimation_mapper.py
+++ b/tests/ops/mapper/test_video_depth_estimation_mapper.py
@@ -5,11 +5,11 @@ import numpy as np
 from data_juicer.core.data import NestedDataset as Dataset
 from data_juicer.ops.mapper.video_depth_estimation_mapper import \
     VideoDepthEstimationMapper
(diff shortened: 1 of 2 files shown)


Qirui-jiao:
Update (November 4):
- Fixed the issue of being unable to import SAM2 due to an outdated Transformers version.
- Fixed the git clone issue for the Video-Depth-Anything repository.
- Updated video_depth_estimation_mapper to support concurrent multi-process execution.

HYLcool:
Skip the test cases for video_depth_estimation_mapper because adding the path to Video-Depth-Anything to the `sys.path` doesn't work in the unittest pipeline, but it works locally. Will check it in the future.

PR #803, opened 2025-11-03
support videos and audios for webdataset & fix custom op modules & optimize video_extract_frames_mapper

Cathy0908:
bugs:
    1. Fixed the bug where modifying SpecialTokens in the configuration file did not take effect in ray mode
    2. Fixed the bug where custom_operator_paths did not take effect in ray mode

new features:
    1. support videos and audios data for webdataset
    2. update video_extract_frames_mapper operator
        Supported output formats are: ["path", "bytes"].
            If format is "path", the output is a list of lists, where each inner
            list contains the path of the frames of a single video.
            e.g.[
                    [video1_frame1_path, video1_frame2_path, ...],
                    [video2_frame1_path, video2_frame2_path, ...],
                    ...
                ] (In the order of the videos).
            If format is "bytes", the output is a list of lists, where each inner
            list contains the bytes of the frames of a single video.
            e.g. [
                    [video1_frame1_bytes, video1_frame2_bytes, ...],
                    [video2_frame1_bytes, video2_frame2_bytes, ...],
                    ...
                ] (In the order of the videos).


data_juicer/ops/mapper/generate_qa_from_examples_mapper.py
```diff
@@ -209,6 +209,9 @@ def parse_output(self, raw_output):
             output_qa_pairs.append((question.strip(), answer.strip()))
         return output_qa_pairs
 
+    def empty_history(self):
+        return []
+
```
HYLcool: Why not use the empty_history method in the parent OP class?

In previous tests, the data type of the empty history should be specified to "str" as well to avoid possible schema alignment error. Thus it returns a empty numpy array with dtype=str.

https://github.com/datajuicer/data-juicer/blob/19c5bb82561db687ba16b50b96526f8f7f4fea4a/data_juicer/ops/base_op.py#L308-L309

Cathy0908: ray would fail using it

HYLcool (approved):

PR #804, opened 2025-11-04
Add vggt_mapper

Qirui-jiao:
- `vggt_mapper`: Input a video of a single scene, and use VGGT to extract information including Camera Pose, Depth Maps, Point Maps, and 3D Point Tracks.


data_juicer/ops/mapper/vggt_mapper.py
```diff
@@ -0,0 +1,221 @@
...
+        vggt_model_path: str = "facebook/VGGT-1B",
+        frame_num: PositiveInt = 3,
+        duration: float = 0,
+        tag_field_name: str = MetaKeys.vggt_tags,
+        frame_dir: str = DATA_JUICER_ASSETS_CACHE,
+        if_output_camera_parameters: bool = False,
+        if_output_depth_maps: bool = False,
+        if_output_point_maps_from_projection: bool = False,
+        if_output_point_maps_from_unprojection: bool = False,
+        if_output_point_tracks: bool = False,
```
HYLcool: It's better to set the default values for these args to True. Otherwise, users applying this OP with the default args might get nothing.

data_juicer/ops/mapper/vggt_mapper.py
```diff
@@ -0,0 +1,221 @@
...
+            with torch.no_grad():
+                # which usually leads to more accurate 3D points than point map branch
+                point_maps_from_unprojection = self.unproject_depth_map_to_point_map(
+                    depth_map.squeeze(0), extrinsic.squeeze(0), intrinsic.squeeze(0)
+                )
+        else:
+            point_maps_from_unprojection = []
+
+        # Predict Tracks
+        query_points = sample.get("query_points")
```
HYLcool, line 192: It's necessary to mention in the doc string that additional query_points are required for predicting the tracks and how to set them (e.g., size, format, normalized or not?).

tests/ops/mapper/test_vggt_mapper.py
HYLcool, line 1: Add some assertions to actually check whether the tests are passed or not. For example, check that if there are expected new fields (e.g., [Fields.meta][self.tag_field_name]["depth_maps"]) in the final dataset and if their formats are expected (e.g., the shapes of depth maps).

data_juicer/ops/mapper/vggt_mapper.py
Qirui-jiao: fixed.

data_juicer/ops/mapper/vggt_mapper.py
Qirui-jiao: fixed.

tests/ops/mapper/test_vggt_mapper.py
Qirui-jiao: fixed.

HYLcool (approved):
LGTM

PR #824, opened 2025-11-19
New Optical Flow OP & Allow to save the computed optical flows

HYLcool:
As the title says.

### DIFF
--- a/data_juicer/ops/filter/video_motion_score_ptlflow_filter.py
+++ b/data_juicer/ops/filter/video_motion_score_ptlflow_filter.py
@@ -50,6 +50,7 @@ class VideoMotionScorePtlflowFilter(VideoMotionScoreFilter):
         self,
         min_score: float = 1.0,
         max_score: float = sys.float_info.max,
+        frame_field: Optional[str] = None,
         model_name: str = "dpflow",
         ckpt_path: Optional[str] = "things",
         get_model_args: Optional[dict] = None,
@@ -67,6 +68,7 @@ class VideoMotionScorePtlflowFilter(VideoMotionScoreFilter):
         super().__init__(
             min_score,
             max_score,
+            frame_field,
             sampling_fps,
             size,
             max_size,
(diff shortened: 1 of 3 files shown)


data_juicer/ops/filter/video_motion_score_ptlflow_filter.py
```diff
@@ -0,0 +1,111 @@
...
+ptlflow = LazyLoader("ptlflow")
+ptlflow_io_adapter = LazyLoader("ptlflow.utils.io_adapter")
+
+OP_NAME = "video_motion_score_ptlflow_filter"
+
+
+@UNFORKABLE.register_module(OP_NAME)
+@OPERATORS.register_module(OP_NAME)
+class VideoMotionScorePtlflowFilter(VideoMotionScoreFilter):
+    """Filter to keep samples with video motion scores within a specified range.
```
yxdyc: Add some comments/remarks about the relationships and comparison of this OP to other two video_motion_xxx OPs

pyproject.toml
```diff
@@ -85,6 +85,8 @@ vision = [
     "rembg",  # Background removal
     "decord", # video/audio handling
     "qwen-vl-utils==0.0.14",  # for Qwen-VL
+    "ptlflow==0.4.1",  # optical flow models collection
```
yxdyc, line 88: The new OP adds many `<` constraints as below
```
   "einops<0.9",
    "h5py<3.13",
    "kaleido<0.3",
    "lightning[pytorch-extra]>=2.1,<2.6",
    "loguru<0.8",
    "opencv-python<4.11",
    "pandas<2.3",
    "plotly<5.25",
    "pypng==0.20220715.0",
    "scipy<1.15",
    "tabulate<0.10",
    "tensorboard<2.19",
    "timm<1.1",
```

Would it be better to merge this OP only after completing the OP-specific environment refactoring?

yxdyc: If so, we can split this PR into two parts and merge the "save the computed optical flows" feature first

HYLcool: I think it's OK. Because: 
1. These constraints will not conflict with current dependencies.
2. If you check deeply, most of these constraints are quite loose, even meaningless. For example, the latest version of loguru is 0.7.x, so the constraint "loguru<0.8" means it accept all available versions of this dependency for now.

PR #837, opened 2025-12-03
rename params name of base op

Cathy0908:
1. rename `cpu_required` to `num_cpus`
2. rename `gpu_required` to `num_gpus`
3. rename `mem_required` to `memory`
4. add arg `runtime_env` for base op
5. add arg `ray_execution_mode` for base op, allow users specifying task or actor mode for each op
6. update `concurency` to `compute` argument (adapt latest ray)


HYLcool:
Passed regression test: https://github.com/datajuicer/data-juicer/actions/runs/20049399758

PR #853, opened 2025-12-16
fix: optimize Ray pipeline performance with column caching and lazy actor initialization

xyuzh:
## Summary

This PR addresses two key issues when running Data-Juicer with Ray:

1. **Pipeline breaking due to repeated `columns()` calls** - Ray's `Dataset.columns()` internally calls `limit(1)` which forces execution and can break streaming pipelines
2. **Premature actor creation before cluster autoscaling** - Deduplicator actors were created at `__init__` time, consuming resources before the cluster had time to scale up

### Column Caching in Ray Dataset (`ray_dataset.py`)
- Cache dataset columns once at the start of `process()` and pass the cached set through the operator pipeline
- Update the cached set when new columns (`Fields.meta`, `Fields.stats`) are added
- Avoids repeated `self.data.columns()` calls that break streaming execution

### Lazy Actor Initialization in Deduplicators
- **`ray_basic_deduplicator.py`**: `ActorBackend` now defers actor creation until first use via `_ensure_actors()`
- **`ray_bts_minhash_deduplicator.py`**: Union-find actors and edge buffers are created lazily in `_ensure_actors()`, called at the start of `run()`
- **`ray_document_deduplicator.py`**: Pass through the new `dedup_set_num` parameter

### Auto-scaling Support
- Added `dedup_set_num="auto"` parameter (default) that calculates optimal actor count from cluster CPU resources (`CPU/2`, minimum 1)
- Calculation happens at actor creation time when cluster has scaled, not at `__init__` time

### Scheduling Strategy (`ray_utils.py`)
- Changed `NodeAffinitySchedulingStrategy` from `soft=False` to `soft=True` to reduce scheduling failures when target nodes are temporarily unavailable

## Testing
- Tested with large-scale FineWeb data processing on Anyscale with autoscaling worker nodes

## Breaking Changes
None - all changes are backward compatible. The `dedup_set_num` parameter defaults to `"auto"` which preserves the previous behavior of using `CPU/2`.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -150,12 +150,20 @@ class RayDataset(DJDataset):
 
         calculate_ray_np(operators)
 
+        # Cache columns once at start to avoid breaking pipeline with repeated columns() calls
+        # Ray's columns() internally does limit(1) which forces execution and breaks streaming
+        cached_columns = set(self.data.columns())
+
         for op in operators:
-            self._run_single_op(op)
+            cached_columns = self._run_single_op(op, cached_columns)
         return self
 
-    def _run_single_op(self, op):
-        if op._name in TAGGING_OPS.modules and Fields.meta not in self.data.columns():
+    def _run_single_op(self, op, cached_columns=None):
+        # Use cached columns to avoid calling self.data.columns() which breaks pipeline
+        if cached_columns is None:
+            cached_columns = set(self.data.columns())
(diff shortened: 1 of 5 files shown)


xyuzh:
Splitting into two focused PRs for easier review:
- #854: fix: cache columns in Ray dataset to avoid breaking streaming pipeline
- #855: feat: lazy actor initialization for Ray deduplicators to support autoscaling

PR #855, opened 2025-12-16
feat: lazy actor initialization for Ray deduplicators to support autoscaling

xyuzh:
## Summary

This PR improves Ray-based deduplicators by lazily initializing actors, allowing the cluster to autoscale before actors consume resources.

## Problem

Previously, deduplicator actors were created at `__init__` time:
- Actors immediately consumed cluster resources
- `ray.cluster_resources().get('CPU')` returned initial (small) cluster size
- Autoscaling couldn't provision additional nodes before actors were created
- Result: suboptimal actor count based on initial cluster state

### Lazy Actor Initialization
- Defer actor creation until first use via `_ensure_actors()` method
- Actors are created in `run()` when the cluster has had time to autoscale
- Resource calculation happens at actor creation time, not `__init__` time

### Auto-scaling Parameter
- Add `dedup_set_num='auto'` parameter (default)
- Calculates optimal actor count from current cluster CPU resources (`CPU/2`, minimum 1)
- Supports explicit integer values for manual control

### Soft Node Affinity
- Changed `NodeAffinitySchedulingStrategy` from `soft=False` to `soft=True`
- Reduces scheduling failures when target nodes are temporarily unavailable

## Files Changed

- `data_juicer/ops/deduplicator/ray_basic_deduplicator.py` - ActorBackend lazy init
- `data_juicer/ops/deduplicator/ray_bts_minhash_deduplicator.py` - Union-find actors lazy init
- `data_juicer/ops/deduplicator/ray_document_deduplicator.py` - Pass through dedup_set_num param
- `data_juicer/utils/ray_utils.py` - Soft node affinity scheduling

## Breaking Changes

None - `dedup_set_num='auto'` preserves the previous behavior of using `CPU/2`.

## Testing

Tested with large-scale FineWeb data processing on Anyscale with autoscaling worker nodes.

### DIFF
--- a/data_juicer/ops/deduplicator/ray_document_deduplicator.py
+++ b/data_juicer/ops/deduplicator/ray_document_deduplicator.py
@@ -1,6 +1,7 @@
 import hashlib
 import string
 from typing import Union
+
 import regex as re
 
 from ..base_op import OPERATORS


Dludora:
Thank you for your contribution! 
Just a quick tip: please try to run pre-commit locally before pushing to avoid style check failures.

PR #862, opened 2025-12-22
Depth seg new op

archernsy:


archernsy:
This PR introduces two new operators:

DepthAnythingv2: For single-frame depth extraction from videos.
Autoseg2: A fully automatic video segmentation tool (based on SAM1 and SAM2). It can automatically segment all content in a video, dividing frames into different objects or regions without requiring manual prompts.
All feedback and suggestions are welcome

HYLcool:
Thanks for your contribution! Before merging this PR, there are several things to be finished:
1. run the pre-commit tool to format the proposed code.
2. resolve the comments from gemini or other reviewers. If one is not valid, just ignore and click "resolve conversation".

BTW, this PR still needs human reviews. Please generously offer some valuable suggestions and comments. @Cathy0908 @Qirui-jiao

Qirui-jiao:
Thank you for your contribution! I suggest moving the `depth_anything_v2` and `sam2` folders to `data_juicer/ops/common/`. Additionally, it would be great if you could wrap the imports in these folders with LazyLoader.

Alternatively, a better approach would be to git clone these folders into DATA_JUICER_ASSETS_CACHE, and then switch the working directory to import the necessary functions (This avoids the need for LazyLoader and keeps the repository lightweight). You can refer to [vggt_mapper](https://github.com/datajuicer/data-juicer/blob/main/data_juicer/ops/mapper/vggt_mapper.py) for an example.

PR #874, opened 2026-01-02
feat(tracer): add trace_keys option for including sample fields in trace output

JohnGiorgi:
> [!IMPORTANT]
> This PR depends on #873 and should be merged after it.

## Summary

Adds a new `trace_keys` configuration option that allows users to specify fields from the original dataset to be included in the tracer output.

## Usage

Via YAML config:
```yaml
open_tracer: true
trace_num: 10
trace_keys:
    - 'id'
```

Or via command line:
```bash
dj-process --config recipe.yaml --open_tracer true --trace_keys '["id"]'
```

You will now see a new field in the tracer output, `"id"`, which is taken straight from the original dataset:

```json
{"id":"abc-123", "original_text":"Hello", "processed_text":"Processed hello"}
```

## Why?

My primary motivation for this feature is to include a unique identifier from the original dataset in the tracer output (e.g. `"id"`). I figured I would keep the feature generic to allow users to include _any_ field they like in the tracer output, hence `trace_keys: list[str]`.

## Changes

- Add `--trace_keys` argument to config parser (type: `List[str]`, default: `[]`)
- Update `Tracer` class to accept and use `trace_keys` parameter
- Update `trace_mapper` to include specified fields when configured
- Pass `trace_keys` from config to `Tracer` in default executor
- Document `trace_keys` in `config_all.yaml`
- Add unit tests for `trace_keys` functionality (single key, multiple keys, missing field, default behavior)


JohnGiorgi:
Thanks for the review @gemini-code-assist! Good catch on the potential key collision issue.

I've addressed this in commit a1d7b4ccc by:
1. Creating an empty dict first
2. Adding `trace_keys` fields first (so they appear at the start of the output)
3. Then adding `original_text` and `processed_text` (which take precedence if there's a collision)

This ensures the trace data remains accurate even if a user mistakenly includes reserved keys in `trace_keys`, and also improves readability by placing the identifier fields first in the output.

JohnGiorgi:
Rebased this PR on top of #873 to use consistent `original_text` key naming. Please merge #873 first, then this PR.

Issue #796, opened 2025-10-14
[Bug]: validator 字段类型校验错误地读取yaml中的 field_types为str 导致字段类型校验的 isinstance 抛出异常


kongzhinvwang2:
### Before Reporting 报告之前

- [x] I have pulled the latest code of main branch to run again and the bug still existed. 我已经拉取了主分支上最新的代码，重新运行之后，问题仍不能解决。

- [x] I have read the [README](https://github.com/alibaba/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process. (Otherwise, we recommend that you can ask a question using the Question template) 我已经仔细阅读了 [README](https://github.com/alibaba/data-juicer/blob/main/README_ZH.md) 上的操作指引，并且在安装过程中没有错误发生。（否则，我们建议您使用Question模板向我们进行提问）

### Search before reporting 先搜索，再报告

- [x] I have searched the Data-Juicer [issues](https://github.com/alibaba/data-juicer/issues) and found no similar bugs. 我已经在 [issue列表](https://github.com/alibaba/data-juicer/issues) 中搜索但是没有发现类似的bug报告。

### OS 系统

Linux

### Installation Method 安装方式

pip

### Data-Juicer Version Data-Juicer版本

latest

### Python Version Python版本

3.10

### Describe the bug 描述这个bug

yaml文件中配置 validator field_type（官方脚本）
`validators:                                                 # validators are a list of validators to be applied when loading a dataset
                                                            # it checks a sample of the dataset for each validator
                                                            # check data_juicer/ore/data/data_validator.py for more validator options
  - type: 'required_fields'                                 # required_fields is a validator to check the required fields in the dataset.
    required_fields:                                          # required_fields is a list of required fields.
      - "text"
    # field_types:                                              # field_types is a dictionary of field types.
    #   text: 'str'`

其中 field_types 在 data_juicer/core/data/data_validator.py 中被设置为`expected_type = self.field_types.get(field)`
这会导致读取到的 expected_type 为字符串类型的 str、list....
在校验时 `invalid_types = [type(v) for v in sample_values if v is not None and not isinstance(v, expected_type)]` 没有将 expected_type 转为 type 类型，导致抛出异常
**TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union**

### To Reproduce 如何复现

只要 validator 的yaml 文件设置 field_types 即可复现

### Logs 报错日志

[...]

PR #886, opened 2026-01-15
 fix: convert string type names to Python types in RequiredFieldsValidator (#796)

dubin555:
fix: convert string type names to Python types in RequiredFieldsValidator (#796)

  PR Description

  ## Summary

  - Fix TypeError when using `field_types` in YAML config for `RequiredFieldsValidator`
  - Add `TYPE_NAME_MAPPING` to convert string type names (e.g., `'str'`, `'int'`) to Python type objects
  - Add `_normalize_field_types()` method to handle both string and type object specifications
  - Add unit tests for string type names, unknown type names, and invalid type specifications

  ## Problem

  When `field_types` is specified in YAML config as documented in `config_all.yaml`:

  ```yaml
  field_types:
    text: 'str'

  The YAML parser reads 'str' as a string "str" rather than Python's str type. This causes isinstance() in the validation logic to raise:

  TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union

  Solution

  Added a type name mapping and normalization method that converts string type names to actual Python types during validator initialization, while maintaining backward compatibility with direct type object specifications.

  Supported type names: str, string, int, integer, float, bool, boolean, list, dict, tuple, set, bytes

  Test plan

  - Verify fix reproduces and resolves the original issue
  - Add unit tests for string type names from YAML config
  - Add unit tests for unknown type name error handling
  - Add unit tests for invalid type specification error handling
  - Ensure backward compatibility with existing type object specifications

  Fixes: https://github.com/datajuicer/data-juicer/issues/796

### DIFF
--- a/data_juicer/core/data/data_validator.py
+++ b/data_juicer/core/data/data_validator.py
@@ -255,18 +255,18 @@ class RequiredFieldsValidator(DataValidator):
 
     # Mapping from type name strings to actual Python types
     TYPE_NAME_MAPPING: Dict[str, type] = {
-        'str': str,
-        'string': str,
-        'int': int,
-        'integer': int,
-        'float': float,
-        'bool': bool,
-        'boolean': bool,
-        'list': list,
-        'dict': dict,
-        'tuple': tuple,
-        'set': set,
-        'bytes': bytes,
+        "str": str,
+        "string": str,
+        "int": int,
+        "integer": int,
+        "float": float,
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM. I'll resolve these minor comments for now and proceed with the merge. If there are any further optimizations, feel free to submit a follow-up PR. Thanks!

PR #892, opened 2026-01-23
[feature] op-level isolated environment spec in ray mode

HYLcool:
- Target: maintain a minimal set of dependencies for core; dynamically install requirements for each OP when running; mainly for ray mode.
- Based on ray runtime environment
- New abstractions for Requirement, Environment Spec for OPs.
- Support to both set the additional requirements manually and analyze them automatically for OPs.
- Support merge possible environments from different OPs that share common dependencies and reuse the created environments.
- Support several resolving strategies when dependency conflicts occur.

### DIFF
--- a/tests/core/data/test_ray_dataset.py
+++ b/tests/core/data/test_ray_dataset.py
@@ -113,11 +113,14 @@ class RayDatasetFuncsTest(DataJuicerTestCaseBase):
     @TEST_TAG('ray')
     def test_get_abs_path_with_nonexistent_local_path(self):
         """Test get_abs_path when local path doesn't exist"""
-        # When the joined path doesn't exist, it should return the original path
-        dataset_dir = "/tmp/nonexistent_dataset"
+        # When the joined path doesn't exist, it should return the current path
+        dataset_dir = "./nonexistent_dataset"
         path = "existing_file.txt"
+        tgt_path = os.path.join(dataset_dir, path)
+        non_tgt_path = os.path.abspath(tgt_path)
         result = self.get_abs_path(path, dataset_dir)
-        self.assertEqual(result, path)
+        self.assertEqual(result, tgt_path)
+        self.assertNotEqual(result, non_tgt_path)
 
 class TestRayDataset(DataJuicerTestCaseBase):
     def setUp(self):


data_juicer/config/config.py
```diff
@@ -478,6 +478,34 @@ def init_configs(args: Optional[List[str]] = None, which_entry: object = None, l
...
             )
+            parser.add_argument(
+                "--open_op_env_manager",
+                type=bool,
+                default=False,
+                help="Whether to open the operator environment manager to automatically analyze and merge runtime "
+                "environment for different OPs. It's only available in ray mode. It helps different OPs share "
+                "the same runtime environment if possible. If it's False, every OP has its own runtime "
+                "environment during processing. If it's True, Data-Juicer tries to merge environments from "
+                "different OPs that have common dependencies to reuse runtime environment. It's False in default.",
```
yxdyc: Can we have a more specific guidance about when to open this feature?

data_juicer/config/config.py
```diff
@@ -478,6 +478,34 @@ def init_configs(args: Optional[List[str]] = None, which_entry: object = None, l
...
+            )
+            parser.add_argument(
+                "--min_common_dep_num_to_combine",
+                type=int,
+                default=-1,
+                help="The minimum number of common dependencies required to determine whether to merge two operation "
+                "environment specifications. If set to -1, it means no combination of operation environments. "
+                "It's -1 in default. Only available when open_op_env_manager is true.",
```
yxdyc: By default, `min_common_dep_num_to_combine = -1` and `open_op_env_manager = False` lead the same outcome. Maybe we only need to expose the former.

data_juicer/ops/op_env.py
```diff
@@ -0,0 +1,593 @@
...
+            extras=list(req.extras),
+            markers=req.marker,
+            url=req.url,
+        )
+    except Exception:
+        logger.error(f"Failed to parse requirement from the requirement string: {req_str}")
+        return None
+
+
+def parse_requirements_list(req_list: List[str]):
```
yxdyc, line 46: To avoid error when constructing dict with None, how about
```
def parse_requirements_list(req_list: List[str]):
    parsed = []
    failed = []
    for s in req_list:
        r = parse_single_requirement(s)
        if r is None:
            failed.append(s)
            continue
        parsed.append(r)
    if failed:
        logger.warning(f"Failed to parse {len(failed)} requirement(s), ignored: {failed}")
    return parsed
```

data_juicer/ops/op_env.py
```diff
@@ -0,0 +1,593 @@
...
+                    max_spec = parse_version_specifier("||".join(max_str))
+                    include_latest = max_spec.include_max
+                    latest_version = max_spec.max
+            else:
+                include_latest = combined.include_max
+                latest_version = combined.max
+            if latest_version:
+                latest_version = (
+                    SpecifierSet(f"=={latest_version}") if include_latest else SpecifierSet(f"<{latest_version}")
+                )
```
yxdyc, line 546: To enhance observability, consider add
```
if latest_version is None:
    logger.warning(
        f"Dependency conflict for {first_req.name or second_req.name}, "
        f"fallback to unpinned version under LATEST strategy: "
        f"{first_req} vs {second_req}"
    )
```

PR #893, opened 2026-01-27
Add Hand Reconstruction op (HaWoR)

Qirui-jiao:
- `video_hand_reconstruction_hawor_mapper`: Use HaWoR and MoGe-2 for hand reconstruction. 
(The unittest is skipped because users need to download MANO_RIGHT.pkl.)

### DIFF
--- a/data_juicer/ops/common/__init__.py
+++ b/data_juicer/ops/common/__init__.py
@@ -9,7 +9,6 @@ from .helper_func import (
     words_augmentation,
     words_refinement,
 )
-from .mano_func import MANO
 from .special_characters import SPECIAL_CHARACTERS
 
 __all__ = [
--- a/docs/Operators.md
+++ b/docs/Operators.md
@@ -269,8 +269,8 @@ All the specific operators are listed below, each featured with several capabili
 | video_face_blur_mapper | 🎬Video 💻CPU 🟢Stable | Mapper to blur faces detected in videos. 映射器模糊在视频中检测到的人脸。 | [info](operators/mapper/video_face_blur_mapper.md) | - |
 | video_ffmpeg_wrapped_mapper | 🎬Video 💻CPU 🟢Stable | Wraps FFmpeg video filters for processing video files in a dataset. 包装FFmpeg视频过滤器，用于处理数据集中的视频文件。 | [info](operators/mapper/video_ffmpeg_wrapped_mapper.md) | - |
 | video_hand_reconstruction_hawor_mapper | 🎬Video 🚀GPU 🟡Beta | Use HaWoR and MoGe-2 for hand reconstruction. 使用HaWoR和MoGe-2进行手部重建。 | - | - |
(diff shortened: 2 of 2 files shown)


Qirui-jiao:
Updates on Feb 3:
- Merge main.
- Updated test.

data_juicer/ops/mapper/video_hand_reconstruction_hawor_mapper.py
```diff
@@ -0,0 +1,526 @@
...
+            )
+
+        self.hawor_detector_path = hawor_detector_path
+        self.frame_num = frame_num
+        self.duration = duration
+        self.tag_field_name = tag_field_name
+        self.frame_dir = frame_dir
+        self.thresh = thresh
+
+    def detect_track(self, imgfiles: list, hand_det_model: ultralytics.YOLO, thresh: float = 0.5) -> tuple:
```
HYLcool: The `ultralytics.YOLO` is used when defining this method, which might trigger the LazyLoader to install ultralytics when importing this OP. Need to check if it's true. If so, maybe a string type annotation is OK (e.g., "YOLO").

Or just remove the type hint of the arg hand_det_model.

data_juicer/ops/mapper/video_hand_reconstruction_hawor_mapper.py
```diff
@@ -0,0 +1,526 @@
...
+            "right_beta_list": right_beta_list,
+            "right_hand_pose_list": right_hand_pose_list,
+            "right_global_orient_list": right_global_orient_list,
+            "right_transl_list": right_transl_list,
+        }
+
+        return sample
+
+
+class MANO(smplx.MANOLayer):
```
HYLcool: The definition of MANO uses the smplx lib, which could trigger the LazyLoader when importing this OP.
Maybe we can rewrite it like the OP sdxl_prompt2prompt_mapper you implemented before: move the class definition in the `common` package and use LazyLoader to import the MANO in this file and model_utils.py.

https://github.com/datajuicer/data-juicer/blob/ed29c1e8fb612e9204ce235bb2160f6a80bd7fc9/data_juicer/ops/mapper/sdxl_prompt2prompt_mapper.py#L15

Then, add simplx and some other required libraries to the list in LazyLoader.check_packages in the init method of this OP.

data_juicer/ops/common/__init__.py
```diff
@@ -9,6 +9,7 @@
     words_augmentation,
     words_refinement,
 )
+from .mano_func import MANO
```
HYLcool: This line is not necessary and can be removed. Or it will trigger the installation of smplx as well.

BTW, please merge the latest main branch and resolve the conflicts in docs/Operators.md.

data_juicer/ops/mapper/video_hand_reconstruction_hawor_mapper.py
Qirui-jiao: fixed.

data_juicer/ops/common/__init__.py
Qirui-jiao: fixed.

HYLcool (approved):

PR #911, opened 2026-02-11
Feat: Support paimon, iceberg, hudi, delta lake, hdfs data source. 

Dludora:
As the title says.

### DIFF
--- a/data_juicer/core/data/load_strategy.py
+++ b/data_juicer/core/data/load_strategy.py
@@ -2,7 +2,7 @@ import fnmatch
 import os
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Dict, Optional, Type
+from typing import Any, Dict, Optional, Type
 
 import datasets
 from jsonargparse import Namespace
@@ -13,7 +13,11 @@ from data_juicer.core.data.config_validator import ConfigValidator
 from data_juicer.download.downloader import validate_snapshot_format
 from data_juicer.format.formatter import unify_format
 from data_juicer.format.load import load_formatter
-from data_juicer.utils.s3_utils import create_pyarrow_s3_filesystem, validate_s3_path
+from data_juicer.utils.s3_utils import (
+    create_pyarrow_s3_filesystem,
+    get_aws_credentials,
+    validate_s3_path,
+)
 
 # based on executor type and data source type, use different
 # data load strategy to product corresponding datasets
@@ -376,6 +380,108 @@ class DefaultModelScopeDataLoadStrategy(DefaultDataLoadStrategy):
         raise NotImplementedError("ModelScope data load strategy is not implemented")
 
 
+@DataLoadStrategyRegistry.register("default", "remote", "hdfs")
+class DefaultHDFSDataLoadStrategy(DefaultDataLoadStrategy):
+    """
+    data load strategy for HDFS datasets for LocalExecutor
+    Uses fsspec-compatible storage_options passed through huggingface datasets
+    """
+
+    CONFIG_VALIDATION_RULES = {
+        "required_fields": ["path"],
+        "optional_fields": ["host", "port", "user", "kerb_ticket", "extra_conf"],
+        "field_types": {"path": str},
+        "custom_validators": {
+            "path": lambda x: x.startswith("hdfs://"),
+        },
+    }
+
+    def _create_hdfs_fs(self):
+        import pyarrow.fs as fs
+
+        host = self.ds_config.get("host", None)
+        port = self.ds_config.get("port", None)
+        if port is not None:
+            port = int(port)
+        user = self.ds_config.get("user", None)
+        kerb_ticket = self.ds_config.get("kerb_ticket", None)
+        extra_conf = self.ds_config.get("extra_conf", None)
+        return fs.HadoopFileSystem(host=host, port=port, user=user, kerb_ticket=kerb_ticket, extra_conf=extra_conf)
+
+    def load_data(self, **kwargs):
+        from urllib.parse import urlparse
+
+        from data_juicer.core.data import NestedDataset
+
+        path = self.ds_config["path"]
+        load_data_np = kwargs.get("num_proc", 1)
+        text_keys = getattr(self.cfg, "text_keys", ["text"])
+
+        file_path = urlparse(path).path
+        file_extension = os.path.splitext(file_path)[1].lower()
+        file_extension_map = {
+            ".json": "json",
+            ".jsonl": "json",
+            ".txt": "text",
+            ".csv": "csv",
+            ".tsv": "csv",
+            ".parquet": "parquet",
+        }
+        data_format = file_extension_map.get(file_extension, "json")
+
+        hdfs = self._create_hdfs_fs()
+
+        try:
+            # Parquet needs a random-access file handle instead of a forward-only stream.
+            if data_format == "parquet":
+                from pyarrow.parquet import read_table
+
+                with hdfs.open_input_file(file_path) as input_file:
+                    arrow_table = read_table(input_file)
+            else:
+                with hdfs.open_input_stream(file_path) as stream:
+
+                    # Use ray.data functions directly with PyArrow filesystem support
+                    # Ray's read functions support filesystem parameter via PyArrow
+                    if data_format in {"json", "jsonl"}:
+                        # For JSON, we need to use read_json_stream with filesystem
+                        import pyarrow.json
+
+                        arrow_table = pyarrow.json.read_json(stream)
+                    elif data_format in {"csv", "tsv"}:
+                        import pyarrow.csv
+
+                        delimiter = "\t" if file_extension == ".tsv" else ","
+                        parse_opts = pyarrow.csv.ParseOptions(delimiter=delimiter)
+                        arrow_table = pyarrow.csv.read_csv(stream, parse_options=parse_opts)
+                    elif data_format == "text":
+                        import pyarrow.csv
+
+                        read_opts = pyarrow.csv.ReadOptions(column_names=["text"])
+                        parse_opts = pyarrow.csv.ParseOptions(delimiter="\0", quote_char=False)
+                        arrow_table = pyarrow.csv.read_csv(stream, read_options=read_opts, parse_options=parse_opts)
+                    else:
+                        raise ValueError(f"Unsupported data format for hdfs: {file_extension}")
+
+            dataset = datasets.Dataset(arrow_table)
+            dataset = NestedDataset(dataset)
+            dataset = unify_format(
+                dataset,
+                text_keys=text_keys,
+                num_proc=load_data_np,
+                global_cfg=self.cfg,
+            )
+
+            return dataset
+
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to load {data_format} data from HDFS path {path}. "
+                f"Ensure Hadoop native libs and configs are available. "
+                f"Error: {str(e)}"
+            )
+
+
 @DataLoadStrategyRegistry.register("default", "remote", "arxiv")
 class DefaultArxivDataLoadStrategy(DefaultDataLoadStrategy):
     """
@@ -543,6 +649,105 @@ class DefaultS3DataLoadStrategy(DefaultDataLoadStrategy):
             )
 
 
+@DataLoadStrategyRegistry.register("ray", "remote", "hdfs")
+class RayHDFSDataLoadStrategy(RayDataLoadStrategy):
+    """
+    data load strategy for HDFS datasets for RayExecutor
+    Uses PyArrow HadoopFileSystem to read from HDFS
+    """
+
+    CONFIG_VALIDATION_RULES = {
+        "required_fields": ["path"],
+        "optional_fields": ["host", "port", "user", "kerb_ticket", "extra_conf"],
+        "field_types": {"path": str},
+        "custom_validators": {
+            "path": lambda x: x.startswith("hdfs://"),
+        },
+    }
+
+    def _create_hdfs_fs(self):
+        import pyarrow.fs as fs
+
+        host = self.ds_config.get("host", None)
+        port = self.ds_config.get("port", None)
+        if port is not None:
+            port = int(port)
+        user = self.ds_config.get("user", None)
+        kerb_ticket = self.ds_config.get("kerb_ticket", None)
+        extra_conf = self.ds_config.get("extra_conf", None)
+        return fs.HadoopFileSystem(host=host, port=port, user=user, kerb_ticket=kerb_ticket, extra_conf=extra_conf)
+
+    def load_data(self, **kwargs):
+        from data_juicer.core.data.ray_dataset import RayDataset
+
+        path = self.ds_config["path"]
+        hdfs_fs = self._create_hdfs_fs()
+
+        logger.info(f"Loading dataset from HDFS: {path}")
+
+        file_extension_map = {
+            ".json": "json",
+            ".jsonl": "json",
+            ".txt": "text",
+            ".csv": "csv",
+            ".tsv": "csv",
+            ".parquet": "parquet",
+            ".npy": "numpy",
+            ".tfrecords": "tfrecords",
+            ".lance": "lance",
+        }
+
+        auto_detect = False
+        data_source = self.ds_config.get("source", None)
+        if data_source is None:
+            auto_detect = True
+        else:
+            suffix = os.path.splitext(data_source)[1]
+            if suffix in file_extension_map:
+                data_format = file_extension_map[suffix]
+            elif "." + data_source in file_extension_map:
+                data_format = file_extension_map["." + data_source]
+            else:
+                auto_detect = True
+
+        if auto_detect:
+            file_extension = os.path.splitext(path)[1]
+            data_format = file_extension_map.get(file_extension, "parquet")
+            logger.info(f"Auto-detected data format: {data_format}")
+        else:
+            logger.info(f"Using specified data format: {data_format}")
+
+        try:
+            import ray.data
+
+            if data_format in {"json", "jsonl"}:
+                from data_juicer.core.data.ray_dataset import read_json_stream
+
+                dataset = read_json_stream(path, filesystem=hdfs_fs)
+            elif data_format == "parquet":
+                dataset = ray.data.read_parquet(path, filesystem=hdfs_fs)
+            elif data_format == "csv":
+                dataset = ray.data.read_csv(path, filesystem=hdfs_fs)
+            elif data_format == "text":
+                dataset = ray.data.read_text(path, filesystem=hdfs_fs)
+            elif data_format == "numpy":
+                dataset = ray.data.read_numpy(path, filesystem=hdfs_fs)
+            elif data_format == "tfrecords":
+                dataset = ray.data.read_tfrecords(path, filesystem=hdfs_fs)
+            elif data_format == "lance":
+                dataset = ray.data.read_lance(path, filesystem=hdfs_fs)
+            else:
+                raise ValueError(f"Unsupported data format for HDFS: {data_format}")
+
+            return RayDataset(dataset, dataset_path=path, cfg=self.cfg)
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to load {data_format} data from HDFS path {path}. "
+                f"Ensure Hadoop native libs and configs are available. "
+                f"Error: {str(e)}"
+            )
+
+
 @DataLoadStrategyRegistry.register("ray", "remote", "s3")
 class RayS3DataLoadStrategy(RayDataLoadStrategy):
     """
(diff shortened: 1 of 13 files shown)


data_juicer/core/data/load_strategy.py
```diff
@@ -376,6 +380,105 @@ def load_data(self, **kwargs):
...
+                # Ray's read functions support filesystem parameter via PyArrow
+                if data_format in {"json", "jsonl"}:
+                    # For JSON, we need to use read_json_stream with filesystem
+                    import pyarrow.json
+
+                    arrow_table = pyarrow.json.read_json(stream)
+                elif data_format == "parquet":
+                    from pyarrow.parquet import read_table
+
+                    arrow_table = read_table(stream)
```
ArdalanM: `arrow_table = read_table(stream)` dit not work from stream.
I had to load directly from `file_path` like this

```
from pyarrow.parquet import read_table
with hdfs.open_input_file(file_path) as f:
    arrow_table = read_table(f)  
```

data_juicer/core/data/load_strategy.py
```diff
@@ -663,3 +868,272 @@ def load_data(self, **kwargs):
...
+
+        path = self.ds_config["path"]
+
+        logger.info(f"Loading Delta Lake table from path: {path}")
+        try:
+            import ray.data
+
+            from data_juicer.utils.model_utils import filter_arguments
+
+            s3_fs = create_pyarrow_s3_filesystem(self.ds_config)
```
cyruszhang, line 1044: **Bug — unconditional S3 filesystem.** `create_pyarrow_s3_filesystem(self.ds_config)` is called regardless of `path` scheme. Delta tables on local disk, `gs://`, `abfs://`, etc. will all get a wasted/incorrect S3 filesystem object passed to `read_delta`. Gate on `path.startswith("s3://")` and pass `filesystem=None` (or omit) otherwise.

data_juicer/core/data/load_strategy.py
```diff
@@ -663,3 +868,272 @@ def load_data(self, **kwargs):
...
+
+            read_kwargs = {**self.ds_config, **kwargs, "filesystem": s3_fs}
+            kwargs = filter_arguments(ray.data.read_delta, read_kwargs)
```
cyruszhang, line 1047: **Bug — `filesystem` can be silently dropped.** `read_kwargs` includes `filesystem=s3_fs`, but then `kwargs = filter_arguments(ray.data.read_delta, read_kwargs)` filters by signature inspection. Across Ray versions, `read_delta`'s public signature may not expose `filesystem`, in which case the S3 fs is dropped. Safer: filter first on a dict that excludes `filesystem`, then assign `kwargs["filesystem"] = s3_fs` afterwards. Also: aliasing the outer `kwargs` parameter is confusing — rename the local.

data_juicer/core/data/load_strategy.py
```diff
@@ -663,3 +868,272 @@ def load_data(self, **kwargs):
...
+                s3_config["aws_secret_access_key"] = catalog_kwargs.pop("s3.secret_access_key")
+            if "s3.session_token" in catalog_kwargs:
+                s3_config["aws_session_token"] = catalog_kwargs.pop("s3.session_token")
+            if "s3.region" in catalog_kwargs:
+                s3_config["aws_region"] = catalog_kwargs.pop("s3.region")
+            if "s3.endpoint" in catalog_kwargs:
+                s3_config["endpoint_url"] = catalog_kwargs.pop("s3.endpoint")
+            aws_access_key_id, aws_secret_access_key, aws_session_token, aws_region = get_aws_credentials(s3_config)
+
+            catalog_kwargs["s3.access_key_id"] = aws_access_key_id
```
cyruszhang, line 916: **Injects `None` into catalog kwargs.** When the user is relying on the default AWS credential chain (IAM role, ~/.aws/credentials) or using a non-S3 catalog (REST/Hive/Glue), `get_aws_credentials(s3_config)` returns `None`s, and these four lines unconditionally write `None` into `catalog_kwargs`. pyiceberg's `load_catalog` will either error or override correct defaults. Only set the keys when the value is truthy.

data_juicer/core/data/load_strategy.py
```diff
@@ -663,3 +868,272 @@ def load_data(self, **kwargs):
...
```
cyruszhang, line 991: Same `None`-injection issue as `DefaultIcebergDataLoadStrategy` — only set keys when values are truthy. Also: this entire 5-key extract/repop block is duplicated verbatim from `DefaultIcebergDataLoadStrategy` (L904–L919). Extract a helper like `_extract_s3_creds_from_iceberg_catalog(catalog_kwargs)`.

data_juicer/core/data/load_strategy.py
```diff
@@ -376,6 +380,108 @@ def load_data(self, **kwargs):
...
+    CONFIG_VALIDATION_RULES = {
+        "required_fields": ["path"],
+        "optional_fields": ["host", "port", "user", "kerb_ticket", "extra_conf"],
+        "field_types": {"path": str},
+        "custom_validators": {
+            "path": lambda x: x.startswith("hdfs://"),
+        },
+    }
+
+    def _create_hdfs_fs(self):
```
cyruszhang, line 399: **Duplication.** `DefaultHDFSDataLoadStrategy._create_hdfs_fs` and `RayHDFSDataLoadStrategy._create_hdfs_fs` (L668) are byte-identical, as are their `CONFIG_VALIDATION_RULES`. Pull both into a mixin (`HDFSConfigMixin`) or add a shared helper `create_pyarrow_hdfs_filesystem(ds_config)` in `data_juicer/utils/s3_utils.py` (alongside the existing S3 helper). Note: `s3_utils.py` already contains an unused `create_filesystem_from_args` that handles this — see comment there.

data_juicer/core/data/load_strategy.py
```diff
@@ -543,6 +649,105 @@ def load_data(self, **kwargs):
...
```
cyruszhang, line 668: Identical to `DefaultHDFSDataLoadStrategy._create_hdfs_fs` (L399). De-duplicate via mixin or shared helper.

data_juicer/core/data/load_strategy.py
```diff
@@ -376,6 +380,108 @@ def load_data(self, **kwargs):
...
+                dataset,
+                text_keys=text_keys,
+                num_proc=load_data_np,
+                global_cfg=self.cfg,
+            )
+
+            return dataset
+
+        except Exception as e:
+            raise RuntimeError(
```
cyruszhang, line 478: Use `raise RuntimeError(...) from e` to preserve the original traceback. As written, the inner exception is stringified into the message and the actual stack is lost — much harder to debug HDFS auth/network errors. Same pattern at L744 (RayHDFS), L941 (DefaultIceberg), L1010 (RayIceberg), L1054 (RayDelta), L1092 (RayHudi), L1138 (RayPaimon).

data_juicer/core/data/load_strategy.py
```diff
@@ -376,6 +380,108 @@ def load_data(self, **kwargs):
...
+        file_extension = os.path.splitext(file_path)[1].lower()
+        file_extension_map = {
+            ".json": "json",
+            ".jsonl": "json",
+            ".txt": "text",
+            ".csv": "csv",
+            ".tsv": "csv",
+            ".parquet": "parquet",
+        }
+        data_format = file_extension_map.get(file_extension, "json")
```
cyruszhang, line 430: Silent `data_format = file_extension_map.get(file_extension, "json")` — unknown extensions become `"json"`, which then fails inside `pyarrow.json.read_json(stream)` with a confusing parse error. Either raise `ValueError(f"Unsupported HDFS file extension: {file_extension}")` here, or hoist the existing `else: raise ValueError(...)` (L463) out of the inner branch so unknown extensions fail fast.

Also: this map treats both `.json` and `.jsonl` as `json`, but PyArrow's `read_json` requires JSONL (one object per line). Single-document `.json` files will fail with a non-obvious error. Worth documenting or handling.

data_juicer/core/data/load_strategy.py
```diff
@@ -663,3 +868,272 @@ def load_data(self, **kwargs):
...
+
+            read_builder = table.new_read_builder()
+            table_scan = read_builder.new_scan()
+            splits = table_scan.plan().splits()
+            table_read = read_builder.new_read()
+            dataset = table_read.to_ray(splits)
+
+            return RayDataset(dataset, dataset_path=path, cfg=self.cfg)
+
+        except ImportError:
```
cyruszhang, line 1133: Other lakehouse strategies (L936, L1004) have an explicit `except ImportError` with an install hint. `RayPaimonDataLoadStrategy` does too — good. But pls verify: the `from pypaimon...` imports happen inside the `try`, and `ImportError` is caught here. Should be fine, but worth a quick test that this branch actually fires when pypaimon isn't installed (vs. being shadowed by the broader `except Exception` below).

data_juicer/core/ray_exporter.py
```diff
@@ -86,7 +94,7 @@ def __init__(
                 self._fernet = load_fernet_key(encryption_key_path)
 
         # Check if export_path is S3 and create filesystem if needed
-        self.s3_filesystem = None
+        self.fs = None
         if export_path.startswith("s3://"):
```
cyruszhang, line 98: **Bug — `NoneType.startswith`.** Init now allows `export_type` alone (good), but `export_path.startswith("s3://")` here will `AttributeError` when `export_path is None`. Guard: `if export_path and export_path.startswith("s3://")`. Same problem at L84 (encryption check) and L211 (`makedirs`).

Add a test for `RayExporter(export_type="iceberg", export_path=None, ...)` to lock this down.

data_juicer/core/ray_exporter.py
```diff
@@ -157,19 +165,7 @@ def _export_impl(self, dataset, export_path, columns=None):
...
-            # Empty dataset with unknown schema - create an empty file
-            logger.warning(f"Dataset is empty, creating empty export file at {export_path}")
-            os.makedirs(os.path.dirname(export_path) or ".", exist_ok=True)
-            with open(export_path, "w"):
-                pass  # Create empty file
-            return
-
-        # Use provided columns or infer from dataset
-        feature_fields = columns if columns else cols
+        feature_fields = dataset.columns() if not columns else columns
```
cyruszhang, line 168: **Regression — empty-dataset handling.** The previous code explicitly handled `dataset.columns()` returning `None` for empty Ray datasets (creating an empty file and returning early). That guard is gone. With an empty dataset:
- `feature_fields = dataset.columns()` → `None`
- `extra_fields.intersection(feature_fields)` → `TypeError: 'NoneType' object is not iterable`

Restore the early-return for `cols is None`, or short-circuit when `dataset.count() == 0`. Please add a regression test (`test_ray_exporter.py`) that calls `_export_impl` with an empty Ray dataset.

cyruszhang (commented):
Thanks for the lakehouse support! A round of review notes — see inline. Highest priority: the Delta unconditional-S3-filesystem and the regressed empty-dataset handling in `RayExporter._export_impl`. The rest are duplication, error-traceback hygiene, and a few correctness concerns around schema inference and the dead `create_filesystem_from_args` helper.

data_juicer/core/ray_exporter.py
```diff
@@ -297,7 +300,180 @@ def write_others(dataset, export_path, **kwargs):
...
+                logger.info(f"Iceberg table {table_identifier} exists. Writing to Iceberg.")
+                use_iceberg = True
+
+            except NoSuchTableError as e:
+                logger.warning(
+                    f"Iceberg target unavailable ({e.__class__.__name__}). Fallback to exporting to {export_path}..."
+                )
+                import pyarrow as pa
+
+                schema = pa.Schema.from_pandas(dataset.limit(1).to_pandas())
```
cyruszhang, line 336: Schema inference via `pa.Schema.from_pandas(dataset.limit(1).to_pandas())` has two problems:
1. **Lossy:** pandas inference flattens binary blobs, nested struct, `Decimal128`, `LargeList` etc. The Iceberg table created here may have a schema that doesn't match later writes.
2. **Empty-dataset trap:** `dataset.limit(1).to_pandas()` returns an empty DataFrame and infers an empty schema → `create_table` builds a useless table.

Prefer Ray's native arrow schema, e.g. `dataset.schema().base_schema` (Ray ≥ 2.6) or `ray.get(dataset.to_arrow_refs()[0]).schema` if available. Same concern at L416 (`write_paimon`).

data_juicer/core/ray_exporter.py
```diff
@@ -297,7 +300,180 @@ def write_others(dataset, export_path, **kwargs):
...
+            table_commit = None
+
+            try:
+                catalog = CatalogFactory.create(catalog_options)
+                try:
+                    table = catalog.get_table(table_identifier)
+                    logger.info(f"Paimon table {table_identifier} exists. Writing to Paimon.")
+                except TableNotExistException:
+                    logger.info(f"Paimon table {table_identifier} does not exist. Creating it before export.")
+                    pa_schema = pa.Schema.from_pandas(dataset.limit(1).to_pandas())
```
cyruszhang, line 416: Same lossy `pa.Schema.from_pandas(dataset.limit(1).to_pandas())` pattern as `write_iceberg` (L336). Use Ray's arrow schema instead.

Also: `Schema.from_pyarrow_schema` vs. `Schema(pa_schema=...)` (L417–L420) — please pin the supported pypaimon version range in a comment, since the kw-arg name has changed across pypaimon releases.

Design nit: `write_iceberg` falls back to file export on **any** error, but `write_paimon` falls back **only on ImportError** (catalog/write errors propagate). Pick one policy. Strict-by-default with an opt-in `iceberg_fallback_to_file` flag is probably safer — silent fallback to file masks misconfigured catalogs and writes data in the wrong place.

data_juicer/core/ray_exporter.py
```diff
@@ -297,7 +300,180 @@ def write_others(dataset, export_path, **kwargs):
...
+            logger.error(f"Iceberg export is unavailable ({e.__class__.__name__}: {e}). Fallback to file export...")
+
+        if use_iceberg:
+            try:
+                filtered_kwargs = filter_arguments(dataset.write_iceberg, export_extra_args)
+                return dataset.write_iceberg(**filtered_kwargs)
+            except Exception as e:
+                logger.error(f"Write to Iceberg failed during execution: {e}. Fallback to json...")
+
+        suffix = os.path.splitext(export_path)[-1].strip(".").lower()
```
cyruszhang, line 355: When the iceberg path falls back, `os.path.splitext(export_path)` runs against what is actually a **table identifier** (e.g. `nessie.warehouse.events`). `splitext` returns `('nessie.warehouse', '.events')`, so the fallback writes JSON-format data to a path literally named `nessie.warehouse.events`. Either require a separate `fallback_export_path` in `export_extra_args`, or refuse to fall back when `export_path` doesn't look like a filesystem path.

data_juicer/core/exporter.py
```diff
@@ -449,6 +457,66 @@ def to_parquet(dataset, export_path, **kwargs):
...
+                catalog.create_table(table_identifier, schema)
+                use_iceberg = True
+            except Exception as e:
+                logger.error(f"Failed to create Iceberg table: {e}. Fallback to exporting to {export_path}...")
+        except Exception as e:
+            logger.error(f"Unexpected error checking Iceberg: {e}. Fallback to exporting to {export_path}...")
+
+        if use_iceberg:
+            try:
+                import daft
```
cyruszhang, line 498: **Side-effect leak.** `import daft` happens *after* `catalog.create_table` (L489). If `daft` isn't installed, the iceberg table has already been created in the catalog when this raises `ImportError`, leaving an empty table behind that the next run won't recreate.

Fix: hoist `import daft` to the top of `to_iceberg` and raise a clear "install daft" error before touching the catalog. Also: confirm whether `daft` is in `pyproject.toml` (optional/extra), since silently requiring it for iceberg export is a hidden dep.

data_juicer/core/exporter.py
```diff
@@ -449,6 +457,66 @@ def to_parquet(dataset, export_path, **kwargs):
...
+            catalog.load_table(table_identifier)
+            logger.info(f"Iceberg table {table_identifier} exists. Writing to Iceberg.")
+            use_iceberg = True
+
+        except NoSuchTableError as e:
+            logger.warning(
+                f"Iceberg target unavailable ({e.__class__.__name__}). Fallback to exporting to {export_path}..."
+            )
+            # Get pyarrow schema from HF Dataset
+            schema = dataset.features.arrow_schema
```
cyruszhang, line 486: `schema = dataset.features.arrow_schema` is taken before the stats/meta column-removal step earlier in `_export_impl`. If `keep_stats_in_res_ds=True`, those columns are baked into the new Iceberg table's schema permanently. Either compute the schema after column removal, or document this as intentional.

data_juicer/core/exporter.py
```diff
@@ -250,14 +254,16 @@ def _export_impl(self, dataset, export_path, suffix, export_stats=True):
                 feature_fields = set(dataset.features.keys())
                 removed_fields = extra_fields.intersection(feature_fields)
                 dataset = dataset.remove_columns(removed_fields)
-            export_method = Exporter._router()[suffix]
+            export_method = Exporter._router().get(suffix, Exporter.to_parquet)
```
cyruszhang, line 257: Nit: `Exporter._router().get(suffix, Exporter.to_parquet)` — the fallback to `to_parquet` is unreachable, since `__init__` already guarded `if self.suffix not in support_dict: raise NotImplementedError`. Use `[suffix]` and drop the implicit-parquet fallback (which would silently mis-format data anyway).

data_juicer/utils/s3_utils.py
```diff
@@ -117,3 +117,30 @@ def validate_s3_path(path: str) -> None:
     """
     if not path.startswith("s3://"):
         raise ValueError(f"S3 path must start with 's3://', got: {path}")
+
+
+def create_filesystem_from_args(path: str, args: Dict[str, Any]):
```
cyruszhang, line 122: **Dead code.** `create_filesystem_from_args` is added but never imported or called anywhere in the diff (`git grep create_filesystem_from_args` returns only this definition). Meanwhile, `RayExporter.__init__` (L96–L117) and the new HDFS strategies (L399, L668) all inline equivalent logic. Either:
- wire this helper into all four sites and delete the inlined code, or
- remove this function.

Adding API surface that nothing calls is the worst of both worlds.

data_juicer/utils/s3_utils.py
```diff
@@ -117,3 +117,30 @@ def validate_s3_path(path: str) -> None:
...
+        fs = create_pyarrow_s3_filesystem(s3_conf)
+        logger.info(f"Detected S3 export path: {path}. S3 filesystem configured.")
+
+    elif path.startswith("hdfs://"):
+        import pyarrow.fs as pa_fs
+
+        hdfs_keys = ["host", "port", "user", "kerb_ticket", "extra_conf"]
+        hdfs_conf = {k: args.pop(k) for k in hdfs_keys if k in args}
+        if "port" in hdfs_conf:
+            hdfs_conf["port"] = int(hdfs_conf["port"])
```
cyruszhang, line 142: If `args` contains `port=None` (rather than omitting the key), `"port" in hdfs_conf` is true and `int(hdfs_conf["port"])` raises `TypeError: int() argument must be a string... not NoneType`. Defensively check truthiness: `if hdfs_conf.get("port") is not None:`.

data_juicer/core/data/load_strategy.py
Dludora: Thanks for the review! This branch will not be shadowed by the subsequent `except Exception as e`, because `except ImportError` comes first and Python matches exception handlers in order. Keeping the `pypaimon` import inside the `try` block also ensures that a missing `pypaimon` dependency is reported through the dedicated ImportError handler with the installation hint.

cyruszhang (commented):
Three more concrete issues I found:

1. Iceberg S3 catalog kwargs are inconsistent. The loaders pop `s3.endpoint` into `s3_config` but never restore it into `catalog_kwargs`, so MinIO/custom S3 catalogs lose the endpoint before `load_catalog(...)`. Also PyIceberg uses hyphenated keys like `s3.access-key-id` / `s3.secret-access-key`, while the implementation checks underscore keys like `s3.access_key_id`; the demo uses the hyphenated form, so real configs can be skipped.

2. `Exporter(export_type="iceberg")` can overwrite the stats export. `_export_impl` derives `stats_file` with `export_path.replace("." + suffix, "_stats.jsonl")`; when `suffix == "iceberg"` and the fallback path is `fallback.jsonl`, this is a no-op, so stats are written to `fallback.jsonl` and then overwritten by fallback data export.

3. Ray Iceberg/Paimon fallback drops the filesystem. `write_iceberg` / `write_paimon` pass `filesystem` directly as `**fallback_kwargs`, but `write_json` / `write_others` only read `kwargs["export_extra_args"]`, so S3 fallback writes lose the configured filesystem.

PR #919, opened 2026-02-25
Add support for json[l].gz, and make ray dataset support reading json…

HunterLine:
1. Add support for json[l].gz
2. Make ray dataset support reading json[l].gz json[l].zst format

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,9 @@ tests/tools/tmp_*/
 tests/ops/deduplicator/chinese_dedup/
 tests/ops/deduplicator/english_dedup/
 
+# temp directory for distributed Ray tests (shared between containers)
+tmp/
+
 
 # perf bench data
 perf_bench_data/
--- a/README.md
+++ b/README.md
@@ -87,6 +87,15 @@ for s in res_ds:
 ## 📰 News
 
 <details open>
+<summary>[2026-02-12] Release v1.5.0: <b>Partitioned Ray Executor, OP-level Env Management, and More Embodied-AI OPs</b></summary>
+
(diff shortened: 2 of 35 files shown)


cyruszhang:
1. please resolve pre-commit errors
2. can we add some test case?

HunterLine:
Yes, I have resolved the pre-commit errors and added some test cases.

data_juicer/format/json_formatter.py
```diff
@@ -9,7 +9,7 @@ class JsonFormatter(LocalFormatter):
     Default suffixes is `['.json', '.jsonl', '.jsonl.zst']`
     """
 
-    SUFFIXES = [".json", ".jsonl", ".jsonl.zst"]
+    SUFFIXES = [".json", ".jsonl", "json.gz", "jsonl.gz", "json.zst", "jsonl.zst"]
```
HYLcool: The file suffixes are parsed from pathlib.Path.suffix/suffixes, which include the dot (".") in the suffixes as well. So the "." should be added before the newly added suffixes like the ones before (".json", ".jsonl", ...)

HYLcool:
```suggestion
    SUFFIXES = [".json", ".jsonl", ".json.gz", ".jsonl.gz", ".json.zst", ".jsonl.zst"]
```

HunterLine: I'm sorry, I will fix this code error. I had noticed this before, but I don't know why I couldn't fix it successfully. I'm really sorry for this.

HYLcool: No need to apologize at all 😊 — tricky but small issues like this are easy to miss, and that's why we review together. Really appreciate the nice work you've put into this PR, and looking forward to your future contribution!

PR #967, opened 2026-04-14
 exclude work_dir from fingerprint hashing without affecting pickling

### LINKED ISSUE #964: [Bug]: use_cache is broken — auto-generated job_id in work_dir poisons fingerprints
### Before Reporting

- [X] I have asked the [Data-Juicer Q&A Copilot](https://datajuicer.github.io/data-juicer/en/main/docs_index.html) (available on Doc Site, DingTalk, or Discord), but the problem still persists.
- [X] I have pulled the latest code of main branch to run again and the bug still existed.
- [X] I have read the [README](https://github.com/datajuicer/data-juicer/blob/main/README.md) carefully and no error occurred during the installation process.

### Search before reporting

- [X] I have searched the Data-Juicer [issues](https://github.com/datajuicer/data-juicer/issues) and found no similar bugs.

### OS

Linux (also reproducible on macOS)

### Installation Method

from source (pip/uv)

### Data-Juicer Version

main branch (post-PR #748)

### Python Version

3.12

### Describe the bug

**`use_cache: true` never produces cache hits between runs.** Every run recomputes all operators from scratch, even with identical input data and config.

#### Root cause

PR #748 introduced `resolve_job_id()`, which auto-generates a `job_id` containing a timestamp + random UUID (e.g. `20260409_155725_d49e30`). This is appended to `work_dir`, and then injected into **every operator** via `op_attrs`:

```python
# config.py, line ~1030
op_attrs = {
    ...
    "work_dir": cfg.work_dir,   # ← contains unique job_id
}
cfg.process = update_op_attr(cfg.process, op_attrs)
```

This sets `self.work_dir` on every operator instance (via `base_op.py` line 361).

When HuggingFace Datasets computes cache fingerprints, it serializes the processing function via `dill.dumps()`. Since the function is a bound method, `dill` serializes `self` (the operator instance), which includes `self.work_dir`. Because `work_dir` changes every run, **every fingerprint is unique per run**, completely defeating the cache.

#### Proof

Dumping the `dill.dumps()` bytes of a wrapped operator function across two runs and diffing them shows a single difference — the timestamped `work_dir` path:

```
Run 1: ...output/20260409_155725_d49e30...
Run 2: ...output/20260409_155736_198798...
```

Everything else in the serialized bytes is identical.

Cache files **are written** (so it appears to work), but on the next run, new fingerprints are computed, new cache files are created, and the old ones are never hit.

### To Reproduce

1. Create a minimal recipe with `use_cache: true`:

```yaml
project_name: 'cache_test'
text_keys: 'target'
use_cache: true
open_tracer: false

process:
  - document_minhash_deduplicator:
      text_key: 'target'
```

2. Run it twice on the same data:

```bash
dj-process --config recipe.yaml --dataset_path data.jsonl --export_path /tmp/output/out.jsonl
dj-process --config recipe.yaml --dataset_path data.jsonl --export_path /tmp/output/out.jsonl
```

3. Observe: the second run re-executes all operators (progress bars, same timing). No cache hits.

4. Inspect the HF cache directory — you'll see **duplicate** cache files with different fingerprints from each run:

```
~/.cache/huggingface/datasets/json/.../cache-<fp_run1>.arrow
~/.cache/huggingface/datasets/json/.../cache-<fp_run2>.arrow  # different fingerprint, same data
```

### Configs

```yaml
project_name: 'cache_test'
text_keys: 'target'
use_cache: true
open_tracer: false

process:
  - document_minhash_deduplicator:
      text_key: 'target'
```

### Logs

_N/A — no errors are raised; the bug is silent._

### Expected Behavior

With `use_cache: true`, the second run of an identical pipeline on identical data should skip already-computed operators by hitting the HuggingFace Datasets cache (matching fingerprints). The second run should complete significantly faster with log messages indicating cache hits rather than re-execution.

### Suggested fix

`work_dir` is an operational concern (logging, checkpoints) and should not affect data-processing fingerprints. Options:

1. **Exclude `work_dir` from the operator's serialized state** by overriding `__getstate__`/`__reduce__` in the `OP` base class to omit `work_dir` during `dill.dumps()`.
2. **Don't inject `work_dir` into `op_attrs`** — instead, pass it through the `run()` call chain (it's already passed to `dataset.process(ops, work_dir=self.work_dir, ...)` in the executor).
3. **Pin the `job_id`** when `use_cache: true` — e.g. derive it deterministically from the config hash instead of using a timestamp.

Option 2 seems cleanest — `work_dir` is only used for tracing/checkpointing (which happens in the executor, not in operator internals), so operators don't need it as instance state.

### Workaround

Pass a fixed `--work_dir` or `--job_id` to keep the path stable across runs:

```bash
dj-process --config recipe.yaml --dataset_path data.jsonl --export_path /tmp/output/out.jsonl --job_id my_fixed_id
```

### Additional Context

This bug was introduced by PR #748 (merged Feb 2026). Before that PR, `work_dir` did not include a per-run `job_id`, so fingerprints were stable across runs.

cyruszhang:
add a _fingerprint_bytes() method to the OP base class that excludes work_dir, _init_args/_init_kwargs, and bound methods from the fingerprint state. Enhance Hasher.hash_default() to use this method when available, including for bound methods passed to generate_fingerprint

Closes #964

### DIFF
--- a/data_juicer/ops/mapper/annotation/annotation_mapper.py
+++ b/data_juicer/ops/mapper/annotation/annotation_mapper.py
@@ -709,7 +709,7 @@ class LabelStudioAnnotationMapper(BaseAnnotationMapper, ABC):
     # for pickling
     def __getstate__(self):
         """Control how the object is pickled"""
-        state = self.__dict__.copy()
+        state = super().__getstate__()
         # Remove unpicklable attributes
         if "client" in state:
             del state["client"]  # Remove Label Studio client
--- a/data_juicer/ops/mapper/annotation/annotation_mapper.py
+++ b/data_juicer/ops/mapper/annotation/annotation_mapper.py
@@ -709,7 +709,7 @@ class LabelStudioAnnotationMapper(BaseAnnotationMapper, ABC):
     # for pickling
     def __getstate__(self):
         """Control how the object is pickled"""
-        state = super().__getstate__()
+        state = self.__dict__.copy()
         # Remove unpicklable attributes
         if "client" in state:
             del state["client"]  # Remove Label Studio client
--- a/data_juicer/ops/base_op.py
+++ b/data_juicer/ops/base_op.py
@@ -301,6 +301,61 @@ class OP(metaclass=OPMetaClass):
     #   2. a string of the path to the requirements.txt file
     _requirements = None
 
+    # Attributes excluded from serialization fingerprinting.
+    # These do not affect data transformation output, so they must not
+    # contribute to cache keys (dill.dumps → Hasher).
+    _NON_FINGERPRINT_ATTRS = frozenset(
+        {
+            # root cause: unique per run (job_id embedded)
+            "work_dir",
+            # parallelism / scheduling — same result regardless of setting
+            "num_proc",
+            "auto_op_parallelism",
+            # error-handling policy
+            "skip_op_error",
+            # resource / Ray scheduling hints
+            "cpu_required",
+            "gpu_required",
+            "mem_required",
+            "num_cpus",
+            "num_gpus",
+            "memory",
+            "runtime_env",
+            "ray_execution_mode",
+            # raw constructor args stashed by OPMetaClass for Ray actor
+            # reconstruction — not data-affecting
+            "_init_args",
+            "_init_kwargs",
+        }
+    )
+
+    def __getstate__(self):
+        """Return state for pickling/dill, excluding execution-only attrs."""
+        return {k: v for k, v in self.__dict__.items() if k not in self._NON_FINGERPRINT_ATTRS}
+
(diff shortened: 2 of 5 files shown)


data_juicer/ops/mapper/annotation/annotation_mapper.py
```diff
@@ -709,7 +709,7 @@ def get_all_annotations(self) -> Dict[int, Dict]:
     # for pickling
     def __getstate__(self):
         """Control how the object is pickled"""
-        state = self.__dict__.copy()
```
cyruszhang, line 712: this is the only mapper that overrides __getstate__ from base op

HYLcool (approved):

PR #968, opened 2026-04-15
[WIP] feat: add persistent custom operator registry

cmgzn:
### Summary

Add a persistent JSON-based registry (`~/.data_juicer/custom_op.json`) so that user-defined custom operators survive across processes without requiring re-registration on every run.

### Motivation

Previously, custom operators had to be re-loaded via config every time a process started. This made it cumbersome to work with reusable custom ops across sessions, scripts, and CLI invocations.

### Changes

- **`data_juicer/utils/custom_op.py`** *(new)* — Core module for persistent custom op management:
  - JSON registry at `~/.data_juicer/custom_op.json` storing source paths keyed by op name
  - `load_persistent_custom_ops()` replays registrations on startup, auto-cleaning stale entries
  - CLI interface: `python -m data_juicer.utils.custom_op {list,register,unregister,reset}`
  - Dynamic module/package loading extracted from `config.py`

- **`data_juicer/utils/registry.py`** — Add `unregister_module()` to `Registry` class

- **`data_juicer/ops/__init__.py`** — Call `load_persistent_custom_ops()` at import time after built-in ops are loaded

- **`data_juicer/config/config.py`** — Replace inline loading logic with a re-export from `custom_op` for backward compatibility

- **`data_juicer/tools/op_search.py`** — Harden `OPRecord` to handle custom ops with non-standard module paths, missing source files, and absent test files

- **`data_juicer/tools/DJ_mcp_granular_ops.py`** — Adapt MCP tooling for enhanced `OPRecord` fields

- **`docs/DeveloperGuide.md`**, **`docs/DeveloperGuide_ZH.md`** — Document the new persistent registration workflow

### Testing

- **`tests/utils/test_custom_op.py`** 
- **`tests/tools/test_op_search.py`**


data_juicer/utils/custom_op.py
```diff
@@ -0,0 +1,532 @@
...
+                if spec is None:
+                    raise RuntimeError(f"Failed to create spec for '{abs_path}'")
+                module = importlib.util.module_from_spec(spec)
+                # register the module first to avoid recursive import issues
+                sys.modules[module_name] = module
+                spec.loader.exec_module(module)
+            except Exception as e:
+                # Clean up partially-initialized module to avoid stale entries
+                sys.modules.pop(module_name, None)
+                raise RuntimeError(f"Error loading '{abs_path}' as '{module_name}': {e}")
```
ShenQianli, line 90: check if we need to rollback `OPERATORS` here

ShenQianli (changes requested):
the persistence model is at the **operator level**, but the real loading unit is the **module/package path** -> unreliable for package-based custom ops, relative imports, and multi-operator modules -> i think persistence should be path-based instead.

data_juicer/utils/custom_op.py
cmgzn: Added `_rollback_operators()` helper. Both file and package loading branches now snapshot `OPERATORS` before loading and roll back on failure.

cmgzn:
> the persistence model is at the **operator level**, but the real loading unit is the **module/package path** -> unreliable for package-based custom ops, relative imports, and multi-operator modules -> i think persistence should be path-based instead.

The registry is now keyed by registration path (file/directory), not operator name. Operator names are always derived at runtime from the live `OPERATORS` registry after loading. This handles package-based custom ops, relative imports, and multi-operator modules correctly.

PR #969, opened 2026-04-20
[WIP] feat(agent): training-ready data recipes, learnable-value mappers, cross-model similarity

yxdyc:
- Add training-data YAMLs (R0 bridge, R1–R3, CPU-only R3), recipes README, TRAIN_DATA_FIELD_GUIDE.

- New mappers: cross-model cohorts (exact | normalized_query | simhash_lsh + match_basis), syslog/harness noise, error taxonomy (string evidence leaves), learnable value scorer, safety gate, distill trajectory, rewrite hints, training card (JSON string; forces turbo map).

- Wire MetaKeys and mapper __init__ exports; extend bad_case / insight / dialog normalize for the training-data stack.

- Add a training-free before/after report (migration matrix, stage retention, hardness/quality ranking checks).

- Stabilize HF Arrow meta typing: tool_success_ratio uses -1.0 when undefined; usage total_tokens always int; bad_case respects ratio >= 0.

- Docs: demos/agent README links; analysis recipe points to R0 + field guide.


data_juicer/ops/mapper/agent_cross_model_pair_mapper.py
```diff
@@ -0,0 +1,396 @@
...
+
+            best_i, best_row = members[best_idx]
+            best_model = self._meta_get(best_row, self.model_meta)
+            best_version = self._meta_get(best_row, self.version_meta)
+            best_quality = self._score_for_row(best_row)
+
+            peer_models = sorted(
+                {str(m) for m in models if m is not None and str(m).strip()},
+            )
+            has_contrast = len(members) >= self.min_group_size and len(peer_models) >= 2
```
ShenQianli: here `has_contrast` only captures data with >=2 models, regardless of version. for example, {model=qwen-plus, version=3.5}, {model=qwen-plus, version=3.6} -> len(peer_models)=1 -> has_contrast=False. feel free to ignore if it is desired:)

PR #972, opened 2026-04-29
fix: unify schema across batches in JSONStreamDatasource to handle null → concrete type evolution

### LINKED ISSUE #936: [Bug] JSONStreamDatasource fixes schema from the first batch and fails on later `null -> concrete type` evolution
# [Bug] JSONStreamDatasource fixes schema from the first batch and fails on later `null -> concrete type` evolution

## Summary

Data-Juicer's custom JSON stream reader appears to lock schema from the first batch in `data_juicer.core.data.ray_dataset.JSONStreamDatasource._read_stream`.

When an early batch infers a nested field as `null` (for example `meta.url = null`) and a later batch introduces a concrete type for the same field (for example `string`), the reader does not reconcile schema across batches and can fail with errors such as:

```text
Couldn't cast array of type ... to null
```

or an Arrow schema mismatch during ingestion.

This is a **correctness + stability** bug in DJ's custom JSON streaming ingestion path, not just a configuration issue.

- Component: `data_juicer.core.data.ray_dataset.JSONStreamDatasource._read_stream`
- Severity suggestion: `P1`
- Impact: pipeline / ingestion stage can fail on otherwise valid mixed-null JSONL

## Environment

- OS: Ubuntu 24.04.2 LTS
- Kernel: Linux 5.15.0-119-generic x86_64
- Python: 3.12
- Ray: 2.54.0
- PyArrow: 23.0.1
- Data-Juicer: 1.5.0
- Reproduced in: standalone script

## Why this looks like a DJ bug

In DJ source, `RayDataset.read_json()` is explicitly marked as a temporary solution for JSON stream reading and has a TODO to replace it with `ray.data.read_json_stream` once available.

In `JSONStreamDatasource._read_stream`, schema starts as `None`, then after the first batch DJ does:

```python
schema = None
while True:
    batch = reader.read_next_batch()
    table = pyarrow.Table.from_batches([batch], schema=schema)
    if schema is None:
        schema = table.schema
    yield table
```

This means the first batch can determine the schema used for later batches. If the first batch infers a field as `null`, later batches with concrete types may fail instead of being reconciled.

## Expected behavior

Valid mixed-null JSONL should remain readable in stream mode.

Examples of acceptable behavior:

- allow `null -> string/int/struct` evolution across batches when the data is otherwise valid;
- reconcile / widen schema across batches;
- or at minimum raise a clearer error that the custom stream path does not support this schema evolution.

## Actual behavior

The custom DJ stream path can fail on inputs where:

- earlier stream batches contain only `null` for a field;
- later batches introduce a concrete type for the same field.

Observed errors include:

```text
Couldn't cast array of type ... to null
```

and related Arrow schema mismatch failures.

## Minimal reproduction

```python
"""Minimal repro for Data-Juicer JSONStreamDatasource schema-lock bug.

Expected behavior on affected DJ versions:
- DJ read_json_stream fails on mixed null/non-null nested field values.
- Ray native read_json succeeds on the same file.
"""

from __future__ import annotations

import json
import tempfile
from pathlib import Path

import pyarrow.json as paj
import ray
from data_juicer.core.data.ray_dataset import read_json_stream


def _write_payload(path: Path) -> None:
    rows = []
    # First batches contain null-only nested field.
    for i in range(30):
        rows.append({"id": i, "meta": {"url": None}})
    # Later batch introduces concrete type.
    rows.append({"id": 999, "meta": {"url": "https://example.com"}})
    with path.open("w", encoding="utf-8") as f:
        for row in rows:
            f.write(json.dumps(row, ensure_ascii=False) + "\n")


def main() -> None:
    ray.init(ignore_reinit_error=True)
    try:
        tmp_dir = Path(tempfile.mkdtemp(prefix="dj_repro_"))
        payload = tmp_dir / "mixed_schema.jsonl"
        _write_payload(payload)
        print("payload =", payload)

        print("\n[DJ read_json_stream]")
        try:
            ds = read_json_stream(
                str(payload),
                override_num_blocks=1,
                # Small block size forces multiple read batches in stream mode.
                read_options=paj.ReadOptions(use_threads=False, block_size=256),
            )
            print("dj count =", ds.count())
            print("dj schema =", ds.schema())
        except Exception as exc:  # noqa: BLE001
            print("dj error =", type(exc).__name__)
            print(exc)

        print("\n[Ray native read_json]")
        ds2 = ray.data.read_json(str(payload), override_num_blocks=1)
        print("ray count =", ds2.count())
        print("ray schema =", ds2.schema())
    finally:
        ray.shutdown()


if __name__ == "__main__":
    main()
```

## Reproduction scripts

1. DJ bug repro (DJ fails, Ray native succeeds on same input):
`tmp/debug/repro_dj_jsonstream_schema_lock.py`

Run:

```bash
conda run -n LLMData --no-capture-output python /Users/markwu/Pycharm/LLMdata/tmp/debug/repro_dj_jsonstream_schema_lock.py
```

2. Ray `_backfill_missing_fields` TypeError repro:
`tmp/debug/repro_ray_backfill_typeerror.py`

Run:

```bash
conda run -n LLMData --no-capture-output python /Users/markwu/Pycharm/LLMdata/tmp/debug/repro_ray_backfill_typeerror.py
```

## Downstream mitigations currently used (workarounds, not root fix)

We currently apply multiple downstream mitigations in our pipeline wrapper to keep production stable:

1. Bypass DJ `read_json_stream` to `ray.data.read_json` when possible, to avoid `JSONStreamDatasource` stream-read failures on mixed-schema JSONL.
2. Auto-infer and sanitize explicit Arrow schema (drop null-only unstable paths, keep known stable fields typed), then pass it through parsing options.
3. Patch BTS deduplicator `map_batches` kwargs to forward per-op `num_cpus`, so Ray task resource requests are honored.
4. Patch `ray.init` caps and tune Ray Data context (`read_op_min_num_blocks`, `min_parallelism`, block sizing, streaming buffer) for stability under large runs.

These mitigations are implemented in:

```text
finance_cleaning/utils/dj_process_capped.py
```

These workarounds unblock production jobs, but they do not fix the underlying schema-handling behavior in DJ's custom JSON stream reader.

## Possible fix directions

1. Avoid fixing schema from the first batch in `JSONStreamDatasource._read_stream`.
2. Reconcile schemas across batches instead of reusing the first inferred schema unchanged.
3. For valid cases such as `null -> string/int`, allow widening / permissive promotion across batches.
4. Prefer the official Ray JSON reading path when equivalent functionality is available, to reduce behavior drift between DJ custom ingestion and Ray native ingestion.

## Additional context

We currently see two distinct classes of failures in this area:

1. **DJ JSON stream schema-lock bug**: first-batch schema fixation causes `cast ... to null` / schema mismatch failures on later concrete values.
2. **Ray Data struct backfill robustness bug**: separate failure mode in `_backfill_missing_fields` for struct reconciliation.

This issue is specifically about the first one: DJ's custom JSON stream ingestion behavior.

fengrui-z:
## Summary

  Fixes #936

  `JSONStreamDatasource._read_stream` locks the schema from the first batch and reuses it for all subsequent batches. When an
  early batch infers a nested field as `null` (e.g. `meta.url = null`) and a later batch introduces a concrete type (e.g.
  `string`), the forced cast from `string` to `null` fails with `ArrowInvalid`.

  This is a **correctness bug** in DJ's custom JSON streaming ingestion path. Ray's native `ray.data.read_json` handles the same input correctly.

  ## Root Cause

  ```python
  # Before: first batch locks schema, all subsequent batches forced to it
  table = pyarrow.Table.from_batches([batch], schema=schema)
  if schema is None:
      schema = table.schema  # locked forever
```
  Fix

  1. Remove the first-batch schema lock — create table without forced schema
  2. Use pyarrow.unify_schemas to merge schemas across batches, allowing null → concrete type promotion
  3. After unification, cast the batch to the unified schema for consistency

```python
  # After: schema evolves across batches
  table = pyarrow.Table.from_batches([batch])
  if schema is None:
      schema = table.schema
  else:
      unified = pyarrow.unify_schemas([schema, table.schema])
      if not unified.equals(schema):
          schema = unified
      table = pyarrow.Table.from_batches([batch], schema=schema)
```

  unify_schemas internally delegates to Arrow C++ UnifyTypes, which promotes null to the concrete type and recursively handles nested structs.

  Test Plan

  - Verify the minimal repro from #936 passes with this fix
  - Verify ray.data.read_json and read_json_stream produce consistent results on mixed-null JSONL
  - Verify no regression on JSONL files with uniform schema
  - Verify no regression on JSONL files with nested structs

  See #936 for the minimal reproduction script.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -427,62 +427,58 @@ class JSONStreamDatasource(_JSONDatasourceBase):
                 raise ValueError(f"Failed to read JSON file: {path}. Error: {e}") from e
             return
 
-        reader = open_json(
-            f,
-            read_options=self.read_options,
-            **self.arrow_json_args,
-        )
-        schema = None
-        buffered_batches = []
-        while True:
-            try:
-                batch = reader.read_next_batch()
-            except pyarrow.lib.ArrowInvalid:
-                yield from self._read_stream_with_schema_unify(path)
-                return
-            except StopIteration:
-                for b in buffered_batches:
-                    yield pyarrow.Table.from_batches([b])
-                return
+        import pyarrow.json as paj
 
-            if schema is None:
-                schema = batch.schema
-                buffered_batches.append(batch)
-            elif schema.equals(batch.schema):
-                buffered_batches.append(batch)
-            else:
-                try:
-                    unified = pyarrow.unify_schemas([schema, batch.schema])
-                except (pyarrow.lib.ArrowInvalid, pyarrow.lib.ArrowTypeError):
-                    yield from self._read_stream_with_schema_unify(path)
-                    return
-                try:
-                    casted = pyarrow.RecordBatch.from_arrays(batch.columns(), schema=batch.schema)
-                    casted = casted.cast(unified)
-                    buffered_batches = [b.cast(unified) for b in buffered_batches]
-                    schema = unified
-                    buffered_batches.append(casted)
-                except (pyarrow.lib.ArrowInvalid, pyarrow.lib.ArrowTypeError):
-                    yield from self._read_stream_with_schema_unify(path)
-                    return
-
-    def _read_stream_with_schema_unify(self, path: str):
-        import json as json_mod
-
-        fs = getattr(self, "_filesystem", None)
-        if fs is None:
-            fs = pyarrow.fs.LocalFileSystem()
+        def _full_file_fallback():
+            """Re-read the entire file with paj.read_json which handles
+            schema inference across all rows in a single pass."""
+            fs = getattr(self, "_filesystem", None)
+            if fs is not None:
+                with fs.open_input_file(path) as f2:
+                    return paj.read_json(
+                        f2,
+                        read_options=self.read_options,
+                        **self.arrow_json_args,
+                    )
+            return paj.read_json(
+                path,
+                read_options=self.read_options,
+                **self.arrow_json_args,
+            )
 
         try:
-            with fs.open_input_stream(path) as f:
-                content = f.read()
-            if isinstance(content, bytes):
-                content = content.decode("utf-8")
-            rows = [json_mod.loads(line) for line in content.strip().split("\n") if line.strip()]
-            if rows:
-                yield pyarrow.Table.from_pylist(rows)
-        except Exception as e:
-            raise ValueError(f"Failed to read JSON file: {path}. Error: {e}") from e
+            reader = open_json(
+                f,
+                read_options=self.read_options,
+                **self.arrow_json_args,
+            )
+            schema = None
+            batches = []
+            schema_evolved = False
+            while True:
+                try:
+                    batch = reader.read_next_batch()
+                except StopIteration:
+                    break
+                batches.append(batch)
+                if schema is None:
+                    schema = batch.schema
+                elif not schema.equals(batch.schema):
+                    schema = pyarrow.unify_schemas([schema, batch.schema])
+                    schema_evolved = True
+            # Yield all batches with consistent schema.
+            # Use .cast() when schema evolved (from_batches does NOT cast).
+            if schema_evolved:
+                for batch in batches:
+                    yield pyarrow.Table.from_batches([batch]).cast(schema)
+            else:
+                for batch in batches:
+                    yield pyarrow.Table.from_batches([batch])
+        except (pyarrow.lib.ArrowInvalid, pyarrow.lib.ArrowTypeError):
+            # PyArrow's streaming reader cannot handle schema evolution
+            # across block boundaries (e.g. null -> string). Fall back to
+            # paj.read_json() which infers schema across the entire file.
+            yield _full_file_fallback()
 
 
 def read_json_stream(
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -400,6 +400,12 @@ if _JSONDatasourceBase is None:
     )
 
 
+class _SchemaEvolutionDetected(Exception):
+    """Internal exception to trigger fallback to buffered schema unification."""
+
+    pass
+
+
 class JSONStreamDatasource(_JSONDatasourceBase):
     """
     A temp Datasource for reading json stream.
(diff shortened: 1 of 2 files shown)


fengrui-z:
Investigated the performance concern. Key findings:

1. **Buffering is unavoidable** — Ray's internal `_combine_tables` → `_align_struct_fields` requires all tables from a single file to have identical schemas. Yielding tables with mixed schemas (e.g., `meta.url: null` then `meta.url: string`) causes `ArrowInvalid: Struct child array #0 does not match type field: null vs string`.

2. **Actual performance impact is minimal** — Ray materializes the full generator output per read task before passing downstream, so latency is unaffected. The memory cost (holding all batches simultaneously) is equivalent to the `paj.read_json()` fallback path.

3. **Current fix**: Buffer batches + `unify_schemas` for the happy path; fall back to `paj.read_json(path)` when PyArrow's streaming reader throws `ArrowInvalid("changed from")` at block boundaries (the #936 case where null → concrete type evolution can't be handled mid-stream).

cyruszhang (commented):
Thanks for tackling #936. I reproduced the PyArrow behavior locally: `open_json` fails on the null -> string transition while `read_json` succeeds, so the correctness issue here is real.

Two things I think we should address before merging:

1. The fallback reopens with `paj.read_json(path, ...)`. `_read_stream` receives `path` after Ray has normalized it against the configured filesystem, so it is not always a standalone local path/URI. When a caller passes a filesystem, this bypasses the already-open stream/filesystem and can fail. I verified this with a `pyarrow.fs.SubTreeFileSystem`: the absolute-path case succeeds, while the filesystem-relative path case fails with `FileNotFoundError: ... schema_evolution.jsonl`. The fallback should reopen through the same filesystem/input source or otherwise preserve filesystem support; please add a filesystem-backed regression test for this path.

2. `batches = []` now buffers every record batch for every file, even when the schema never evolves. That changes `read_json_stream` from streaming to whole-file memory use. Ray's `ReadTask` yields blocks from the iterator and its datasource contract explicitly allows a single large file to return multiple blocks to avoid OOM, so this can regress large JSONL workloads. If full-file inference is unavoidable for evolving schemas, I think we should make that behavior explicit and/or limit it to the fallback path rather than silently applying it to all reads.

cyruszhang:
> Thanks for tackling #936. I reproduced the PyArrow behavior locally: `open_json` fails on the null -> string transition while `read_json` succeeds, so the correctness issue here is real.
> 
> Two things I think we should address before merging:
> 
> 1. The fallback reopens with `paj.read_json(path, ...)`. `_read_stream` receives `path` after Ray has normalized it against the configured filesystem, so it is not always a standalone local path/URI. When a caller passes a filesystem, this bypasses the already-open stream/filesystem and can fail. I verified this with a `pyarrow.fs.SubTreeFileSystem`: the absolute-path case succeeds, while the filesystem-relative path case fails with `FileNotFoundError: ... schema_evolution.jsonl`. The fallback should reopen through the same filesystem/input source or otherwise preserve filesystem support; please add a filesystem-backed regression test for this path.
> 2. `batches = []` now buffers every record batch for every file, even when the schema never evolves. That changes `read_json_stream` from streaming to whole-file memory use. Ray's `ReadTask` yields blocks from the iterator and its datasource contract explicitly allows a single large file to return multiple blocks to avoid OOM, so this can regress large JSONL workloads. If full-file inference is unavoidable for evolving schemas, I think we should make that behavior explicit and/or limit it to the fallback path rather than silently applying it to all reads.

  One more, independent of the buffering discussion: the unify_schemas happy path is currently broken.
 
  Table.from_batches([batch], schema=unified) does not cast — it requires the batch schema to already equal the target. The moment a batch's schema differs from the unified one, it raises ArrowInvalid: Schema at index 0 was different. Note that message does not contain "changed from", so it escalates straight
  to the ValueError instead of recovering.
  
  import pyarrow as pa
  b1 = pa.RecordBatch.from_pylist([{"meta": {"url": None}}])      # struct<url: null>
  b2 = pa.RecordBatch.from_pylist([{"meta": {"url": "x"}}])       # struct<url: string>
  unified = pa.unify_schemas([b1.schema, b2.schema])             # struct<url: string>

  pa.Table.from_batches([b1], schema=unified)   # ArrowInvalid: Schema at index 0 was different
  pa.Table.from_batches([b1]).cast(unified)     # OK ✅  (pyarrow 23.0.1)

  So the for batch in batches: yield Table.from_batches([batch], schema=schema) loop fails on any batch whose schema isn't already the unified one. #936 never reaches it (it errors earlier inside read_next_batch, via the "changed from" fallback), so this branch is effectively untested dead code today.

  Fix is Table.from_batches([batch]).cast(schema).

cyruszhang:
LGTM

cyruszhang (approved):

PR #973, opened 2026-05-01
Update num_proc handling for vllm and Ray mode

ArdalanM:
### Summary                                                                                                                                                                                                                                       
   
  `TextTaggingByPromptMapper` was unconditionally setting `self.num_proc = 1`, which in a                                                                                                                                                           
  Ray+vLLM run capped the actor pool to a single GPU. The only available multi-GPU strategy
  was tensor parallelism (`tensor_parallel_size=N`); data parallelism was silently broken.                                                                                                                                                          
                                                                                                                                                                                                                                                    
  This fix makes `num_proc = 1` conditional on the execution backend, so Ray can now schedule                                                                                                                                                       
  one vLLM actor per GPU — enabling both data parallelism, tensor parallelism, or a                                                                                                                                                                 
  combination of both.

### DIFF
--- a/data_juicer/ops/mapper/text_tagging_by_prompt_mapper.py
+++ b/data_juicer/ops/mapper/text_tagging_by_prompt_mapper.py
@@ -114,7 +114,11 @@ class TextTaggingByPromptMapper(Mapper):
         """  # noqa
 
         super().__init__(*args, **kwargs)
-        self.num_proc = 1
+        if not enable_vllm or not is_ray_mode():
+            # In Ray mode with vllm, num_proc is left to the caller so Ray can
+            # schedule one actor per GPU for data parallelism.
+            self.num_proc = 1
+            
 
         self.prompt = prompt
         self.tag_list = tag_list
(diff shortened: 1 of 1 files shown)


cmgzn:
Thanks for catching this, @ArdalanM — you've spotted a real bug; the unconditional `self.num_proc = 1` does break data parallelism under Ray + vLLM. 🙏

One thing worth noting: the current `not enable_vllm or not is_ray_mode()` still forces the HuggingFace backend into a single process even on Ray. To stay consistent with the other LLM mappers (`optimize_qa_mapper`, `generate_qa_from_text_mapper`, etc.) and let both backends parallelize, the simplest fix is to just **remove the top-level `self.num_proc = 1`** and keep the existing `if enable_vllm: ... if not is_ray_mode():` branch — a one-line change.

Happy to re-review whenever works for you. If you're short on time, we're glad to pick it up and keep you credited as the original author.

ArdalanM:
Just removed the `self.num_proc = 1`
Thanks !

cmgzn (approved):
LGTM! thanks for your pr~!

PR #978, opened 2026-05-14
Fix Ray deduplicator shared state

### LINKED ISSUE #971: RayBasicDeduplicator 懒加载策略导致无法实现全局去重
在 ray_dataset.py 里，算子通过 dataset.map_batches(op.compute_stats, ...) 分发到各个 worker。Ray 在调度时会对 op.compute_stats（bound method）进行 pickle 序列化，而此时 op.backend._dedup_sets = None（因为懒初始化，还没有创建 actor）。每个 worker 进程都独立创建自己的一套 actor，各 worker 的 DedupSet 互相隔离

实际后果： 不同 worker 处理的 batch 之间无法共享去重状态，跨 worker 的重复文档无法被检出，导致去重不完整。

macroguo-ghy:
## Summary

- Share Ray deduplicator backend state across `map_batches` tasks for a single execution.
- Prepare Ray actor-backed dedup sets before serializing the operator into Ray tasks without recreating existing actor handles.
- Materialize Ray basic deduplicator stats for all stateful backends, including Redis, before later actions can re-run the lazy stats stage.
- Add regression coverage for document deduplication across Ray blocks, repeated executions, Redis materialization signaling, and actor handle reuse.
- Fix Ray test helper conversion by using `RayDataset.to_list()` instead of iterating `RayDataset` directly.

Fixes #971.

## Validation

```bash
python3 -m pytest tests/ops/deduplicator/test_ray_document_deduplicator.py -q
```

Result: `9 passed, 10 warnings in 69.43s`.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -277,6 +277,11 @@ class RayDataset(DJDataset):
                     cached_columns.add(Fields.stats)
                 prepare_for_ray_map_batches = getattr(op, "_prepare_for_ray_map_batches", None)
                 use_instance_for_ray_tasks = bool(prepare_for_ray_map_batches and prepare_for_ray_map_batches())
+                if use_instance_for_ray_tasks and op.use_ray_actor():
+                    logger.info(
+                        f"{op._name}: overriding ray_execution_mode from actor to task "
+                        f"to preserve shared dedup state across workers"
+                    )
                 if op.use_ray_actor() and not use_instance_for_ray_tasks:
                     compute = get_compute_strategy(op.__class__, concurrency=op.num_proc)
                     self.data = self.data.map_batches(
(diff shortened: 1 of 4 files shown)


fengrui-z:
## Strength

  - **Correct direction.** Eagerly creating dedup actors on the driver and letting Ray pickle the handles to every worker is the canonical pattern for shared stateful operators — the fix targets the actual root cause.
  
  ## Risks
  
  1. **Loss of lazy autoscale.** Actor count is now locked in at planning time based on current `cluster_resources()`, breaking `ActorBackend`'s original deferred-creation design.
  
  2. **`materialize()` buffers the full post-dedup dataset into the object store.** Previously streaming — a real memory cost on large datasets worth noting in the PR description.
  
  3. **Silently overrides `ray_execution_mode='actor'` for dedup ops.** Downgrades to task mode without a log line; users may wonder why their config "didn't take effect."
  
  4. **Hook is informal.** Underscore-prefixed and `getattr`-probed instead of declared on `Filter`. Consider promoting to a formal `Filter.prepare_for_ray_map_batches() -> bool` API for future reuse.
  
  5. **Repeated-read regression test depends on Ray's scheduling** and didn't reproduce the original failure locally. A unit test that directly asserts `materialize()` was called would be more robust across Ray versions.
  
  ## Recommendation
  
  **Approve and merge** after noting the `materialize()` memory cost and lost lazy-autoscale in the PR description. Optionally: promote the hook to a formal `Filter` API.

Qirui-jiao (approved):

PR #979, opened 2026-05-14
fix(checkpoint): resolve relative path issue in Ray checkpoint writer

cmgzn:
Ray's write_parquet is executed by Ray workers whose working directory may differ from the main process. When checkpoint_dir or work_dir is a relative path (e.g., './tmp/...'), Ray workers cannot resolve it correctly, resulting in empty checkpoint directories.

Fix: convert work_dir and checkpoint_dir to absolute paths during PartitionedRayExecutor initialization.


yxdyc (approved):
Do we have unit tests for the ray-mode path?

cmgzn:
> Do we have unit tests for the ray-mode path?

Good catch — we didn't have ray-mode path tests before. Just added them covering relative→absolute conversion, URI preservation, and end-to-end validation. All passing.

PR #980, opened 2026-05-15
use custom gzip compression to tolerate FUSE-mounted OSS/S3

ideal:
Loguru's built-in "gz" compression fails on K8s + Fluid OSS/S3 mounts because the post-compression unlink may returns OSError errno 34 even though the source file is removed. We replace it with a custom callback that swallows that specific failure so log compression works under FUSE.

e.g.:
image

### DIFF
--- a/data_juicer/utils/logger_utils.py
+++ b/data_juicer/utils/logger_utils.py
@@ -153,14 +153,7 @@ def get_log_file_path():
             return handler._sink._file.name
 
 
-def setup_logger(
-    save_dir,
-    distributed_rank=0,
-    filename="log.txt",
-    mode="o",
-    level="INFO",
-    redirect="auto",
-):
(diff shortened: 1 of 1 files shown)


cmgzn:
Thanks for the detailed fix and context.

I agree this is a real issue, but the root cause appears to be the FUSE/Fluid mount returning `ERANGE` from `unlink`. I’m hesitant to replace Loguru’s built-in `compression="gz"` globally in Data-Juicer for a filesystem-specific behavior, since that would diverge from Loguru’s native compression semantics.

On the Data-Juicer side, I’m a bit concerned that swallowing filesystem errors here (even if limited to `ERANGE`) isn’t the cleanest/most reliable long-term approach, since it’s hard to be confident we’re not hiding a real deletion failure in some cases.

I’d prefer to address this in the FUSE/Fluid layer if possible.

PR #981, opened 2026-05-19
[codex] fix initialize_ray without config

macroguo-ghy:
## Summary
- avoid reading cfg.get when initialize_ray is called with cfg=None
- keep custom_operator_paths behavior unchanged when a config is provided
- add focused unit coverage for no-config initialization and custom operator paths

## Root Cause
initialize_ray already defaulted the Ray address to auto when cfg was None, but runtime_env construction still dereferenced cfg.get, raising AttributeError before ray.init could run.

## Tests
- python3 -m unittest tests.utils.test_ray_utils
- python3 -m py_compile data_juicer/utils/ray_utils.py tests/utils/test_ray_utils.py
- git diff --check


cmgzn (approved):
LGTM, thanks for the pr!

PR #984, opened 2026-05-25
feat: Wire override_num_blocks through full call chain for Ray Data read operations

fengrui-z:
### Summary

Enable users to control Ray Data's block parallelism via a new `--override_num_blocks` CLI argument. This parameter was already implemented at the lowest layer (`read_json_stream()`) but was never wired through the upstream call chain, making it inaccessible without monkey-patching.

### Motivation

When processing very large datasets (billions of records / PB-scale), Ray Data's default block size (128MB) creates an excessive number of blocks (~40M blocks for 5PB), leading to:

- **Driver OOM**: Block metadata (~1KB/block) overwhelms driver memory
- **Poor parallelism control**: Users cannot tune the read parallelism to match their cluster topology (e.g., 96 CPUs need different block counts than 8 CPUs)
- **Startup latency**: Default partitioning scans all blocks before processing begins

Previously, the only workaround was to monkey-patch `RayLocalJsonDataLoadStrategy.load_data()` (as seen in benchmark scripts). This PR provides a clean, user-facing configuration path.

### Changes

| File | Change |
|------|--------|
| `data_juicer/config/config.py` | Add `--override_num_blocks` CLI argument (`Optional[int]`, default `None`) |
| `data_juicer/core/executor/ray_executor_partitioned.py` | Extract `cfg.override_num_blocks` and pass to `load_dataset()` |
| `data_juicer/core/data/load_strategy.py` | `RayLocalJsonDataLoadStrategy.load_data()` extracts and forwards `override_num_blocks` |
| `data_juicer/core/data/ray_dataset.py` | `read()` accepts `**kwargs`; `read_json()` accepts `override_num_blocks` and passes to `read_json_stream()` |

### Usage

```bash
# Control block parallelism for large datasets
python process.py --override_num_blocks 1000 --dataset_path /data/huge_dataset.jsonl

# Or in YAML config
override_num_blocks: 1000
```

### Backward Compatibility

- Default is `None` (no override) — existing behavior is unchanged
- All changes use `**kwargs` pass-through or `Optional` parameters with `None` defaults
- No breaking changes to existing APIs

### Testing

- Smoke test: `python -c "import data_juicer"` passes
- The parameter correctly flows through: `cfg` → `PartitionedRayExecutor._run_impl()` → `DatasetBuilder.load_dataset()` → `RayLocalJsonDataLoadStrategy.load_data()` → `RayDataset.read()` → `RayDataset.read_json()` → `read_json_stream()` → `ray.data.read_datasource(override_num_blocks=N)`

### Related

- Prerequisite for PB-scale data processing support
---


cmgzn:
From the current config surface, this looks like the first top-level config that maps specifically to a `ray.data.read_*` argument. Since Ray read APIs expose quite a few related options (`override_num_blocks`, `concurrency`, `ray_remote_args`, `shuffle`, etc.), I wonder if it would be better to introduce a generic Ray read config shape instead of adding a one-off top-level field.

For example:

```yaml
ray_read_options:
  override_num_blocks: 1000
  concurrency: 64
  ray_remote_args:
    num_cpus: 0.25
```

fengrui-z:
> From the current config surface, this looks like the first top-level config that maps specifically to a `ray.data.read_*` argument. Since Ray read APIs expose quite a few related options (`override_num_blocks`, `concurrency`, `ray_remote_args`, `shuffle`, etc.), I wonder if it would be better to introduce a generic Ray read config shape instead of adding a one-off top-level field.
> 
> For example:
> 
> ```yaml
> ray_read_options:
>   override_num_blocks: 1000
>   concurrency: 64
>   ray_remote_args:
>     num_cpus: 0.25
> ```


Good suggestion — I considered this but decided to keep a top-level argument for the following reasons:

1. **Type safety & validation.** A dedicated `--override_num_blocks` (type `Optional[int]`) catches typos and type errors at parse time. A generic dict silently ignores misspelled keys (e.g. `overide_num_blocks: 1000` → no error, no effect).

2. **CLI discoverability.** A top-level arg shows up in `--help` and supports tab-completion. With a dict, users need to know the internal key names upfront:
   ```bash
   # Clear and discoverable
   python process.py --override_num_blocks 1000
   
   # Requires prior knowledge of key names
   python process.py --ray_read_options '{"override_num_blocks": 1000}'
   ```

3. **The dict would be immediately destructured.** Tracing through the call chain, the dict gets `**`-spread at the executor and each key is individually popped in `RayDataset.read()`. It never operates *as a dict* — it's purely a transport wrapper, adding indirection with no functional benefit.

4. **YAGNI.** `override_num_blocks` is currently the **only** Ray read param that requires upstream wiring. The other candidates (`concurrency`, `ray_remote_args`) are auto-tuned by Ray and rarely need user override. If/when a second param genuinely needs to be exposed, that's a natural point to refactor into a grouped config.

5. **Precedent in this codebase.** `--ray_address`, `--batch_size`, and other important Ray-related params are all top-level explicit args — grouping just this one into a dict would be inconsistent.

Happy to revisit if we find ourselves adding a second or third Ray read param in the near future.

thirdparty/LLM_ecosystems/setup_helm.sh
cmgzn, line 1: I noticed these files only changed executable mode (`100755` -> `100644`) with no content changes. Since they have shebangs, I’m not sure if that was intentional.

data_juicer/core/data/ray_dataset.py
```diff
@@ -369,23 +372,35 @@ def read(cls, data_format: str, paths: Union[str, List[str]]) -> RayDataset:
                 from data_juicer.utils.lazy_loader import LazyLoader
 
                 LazyLoader.check_packages(["pylance"])
-            return getattr(ray.data, f"read_{data_format}")(paths)
+            dataset = getattr(ray.data, f"read_{data_format}")(paths, **kwargs)
```
cmgzn: I checked this against the repo’s minimum Ray version (`ray[default]>=2.51.0`). These native Ray readers already accept `override_num_blocks` directly, so we should pass it into the `ray.data.read_*` call instead of applying `repartition()` afterward.

One scope question: should the regular Ray executor (`data_juicer/core/executor/ray_executor.py`) and the Ray S3 loader (`data_juicer/core/data/load_strategy.py`) also forward `override_num_blocks`, so this option is not limited to the `ray_partitioned` local-read path?

cmgzn:
> > From the current config surface, this looks like the first top-level config that maps specifically to a `ray.data.read_*` argument. Since Ray read APIs expose quite a few related options (`override_num_blocks`, `concurrency`, `ray_remote_args`, `shuffle`, etc.), I wonder if it would be better to introduce a generic Ray read config shape instead of adding a one-off top-level field.
> > For example:
> > ```yaml
> > ray_read_options:
> >   override_num_blocks: 1000
> >   concurrency: 64
> >   ray_remote_args:
> >     num_cpus: 0.25
> > ```
> 
> Good suggestion — I considered this but decided to keep a top-level argument for the following reasons:
> 
> 1. **Type safety & validation.** A dedicated `--override_num_blocks` (type `Optional[int]`) catches typos and type errors at parse time. A generic dict silently ignores misspelled keys (e.g. `overide_num_blocks: 1000` → no error, no effect).
> 2. **CLI discoverability.** A top-level arg shows up in `--help` and supports tab-completion. With a dict, users need to know the internal key names upfront:
>    ```shell
>    # Clear and discoverable
>    python process.py --override_num_blocks 1000
>    
>    # Requires prior knowledge of key names
>    python process.py --ray_read_options '{"override_num_blocks": 1000}'
>    ```
> 3. **The dict would be immediately destructured.** Tracing through the call chain, the dict gets `**`-spread at the executor and each key is individually popped in `RayDataset.read()`. It never operates _as a dict_ — it's purely a transport wrapper, adding indirection with no functional benefit.
> 4. **YAGNI.** `override_num_blocks` is currently the **only** Ray read param that requires upstream wiring. The other candidates (`concurrency`, `ray_remote_args`) are auto-tuned by Ray and rarely need user override. If/when a second param genuinely needs to be exposed, that's a natural point to refactor into a grouped config.
> 5. **Precedent in this codebase.** `--ray_address`, `--batch_size`, and other important Ray-related params are all top-level explicit args — grouping just this one into a dict would be inconsistent.
> 
> Happy to revisit if we find ourselves adding a second or third Ray read param in the near future.

Thanks, that rationale makes sense to me.

thirdparty/LLM_ecosystems/setup_helm.sh
fengrui-z: Not intentional — my local git config was stripping the executable bit. Restored in the latest push. Thanks for catching it.

cmgzn (approved):
LGTM~

PR #988, opened 2026-05-30
fix(core): add FUSE-safe rmtree fallback to PartitionedRayExecutor

wenhaozhao011-cmd:
PR #943 fixed shutil.rmtree() failures on FUSE-mounted OSS buckets (Fluid) in RayExecutor, but missed the same pattern in ray_executor_partitioned.py. When shutil.rmtree() fails with OSError (e.g. errno 34 on FUSE devices), fall back to os.rmdir() which succeeds in this scenario.

Applies the fix to all 3 shutil.rmtree sites:
- TempDirManager.__exit__
- cleanup_temp_files()
- _clear_invalid_checkpoints()

### DIFF
--- a/data_juicer/core/executor/ray_executor_partitioned.py
+++ b/data_juicer/core/executor/ray_executor_partitioned.py
@@ -55,10 +55,7 @@ class TempDirManager:
             try:
                 shutil.rmtree(self.tmp_dir)
             except OSError as e:
-                logger.warning(
-                    f"Remove tmp dir with shutil.rmtree() failed: {e}, "
-                    "will try os.rmdir()"
-                )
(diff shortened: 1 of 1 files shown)


cmgzn (changes requested):
`pre-commit` is currently failing on this PR because Black reformats `data_juicer/core/executor/ray_executor_partitioned.py`.

The CI log shows Black changing the three newly added multi-line `logger.warning(...)` calls, and I reproduced it locally with:

```bash
uv run --with black black --check data_juicer/core/executor/ray_executor_partitioned.py
```

Please run Black/pre-commit on this file and push the formatting update so the check can pass.

cmgzn (approved):
LGTM! thanks for your pr!

PR #992, opened 2026-06-18
fix(op_env): resolve LATEST strategy returning unpinned version for unbounded ranges

CodyQin:
## Summary

Fixes a bug in `OPEnvManager._resolve_with_strategy()` where the LATEST conflict
resolution strategy incorrectly falls back to an unpinned (unconstrained) version
when the union of two conflicting version specifiers contains ranges without an
upper bound.

## Problem

When resolving conflicts like `numpy>=2.0` vs `numpy<1.5`, the union produces
`<1.5||>=2.0` (a UnionSpecifier with one unbounded range). The original code
breaks out of the loop when encountering any range with `max=None`, discarding
all version information and returning a Requirement with no version constraint.

This allows pip to install any version, potentially breaking both OPs.

## Fix

When a UnionSpecifier contains unbounded ranges (`max=None`), instead of falling
back to unpinned, we now use the highest lower bound (`min`) from those unbounded
ranges as the version constraint:

- `>=2.0 | <1.5` → `>=2.0` (highest min from unbounded ranges)
- `>1.23.0 | <1.22.0` → `>1.23.0`

Only when no lower bound exists at all (fully open range) does the fallback to
unpinned + warning remain.

## Tests

- Updated `test_conflict_resolution_latest_strategy_without_max` to reflect corrected behavior
- Added `test_conflict_resolution_latest_incompatible_uses_highest_min`
- Added `test_conflict_resolution_latest_multiple_unbounded_picks_max_min`
- Added `test_conflict_resolution_latest_unbounded_include_min_false`

All 60 tests in `test_op_env.py` pass.

### DIFF
--- a/data_juicer/ops/op_env.py
+++ b/data_juicer/ops/op_env.py
@@ -523,28 +523,43 @@ class OPEnvManager:
             # use union to combine them to find the latest version
             combined = p1 | p2
             if isinstance(combined, UnionSpecifier):
-                max_str = []
-                for r in combined.ranges:
-                    if r.max:
+                has_unbounded = any(r.max is None for r in combined.ranges)
+                if has_unbounded:
+                    # Some ranges extend to +inf. Use the highest lower bound
+                    # from unbounded ranges as the LATEST constraint.
+                    max_min = None
+                    include_max_min = False
+                    for r in combined.ranges:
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM! Thanks for the pr.

PR #994, opened 2026-06-23
fix: update deprecated models and skip image_sam_3d_body_mapper test

cmgzn:
### 1. Update deprecated model names

Replace deprecated model names with available ones across test files, demo configs, and docstrings:

| File | Old Model | New Model |
|---|---|---|
| `text_chunk_mapper.py` docstring | `qwen2.5-72b-instruct` | `qwen3.7-max` |
| `role_playing_system_prompt.yaml` | `qwen2.5-72b-instruct` | `qwen3.7-max` |
| `system_prompt_generator.py` | `qwen2.5-72b-instruct` | `qwen3.7-max` |
| `test_image_tagging_vlm_mapper.py` | `qwen2.5-vl-3b-instruct` | `qwen3-vl-8b-instruct` |

### 2. Skip `image_sam_3d_body_mapper` test (temporary)

The `image_sam_3d_body_mapper` operator installs MoGe and its dependencies via `pip install` during initialization, which pollutes the global Python environment (e.g. upgrading `gradio`, `huggingface-hub`, `fastapi`, `starlette`) and causes regression test failures for other operators (such as vLLM-related ops).

A previous attempt constrained MoGe's runtime dependencies during installation, but this approach was too ad-hoc — the same issue would recur whenever MoGe or other dependencies are updated. After discussion, it was decided to:

- **Temporarily skip all tests for this operator** to avoid polluting global dependencies.
- Discuss a fundamental solution (e.g. isolated runtime dependency management) separately.

No changes were made to `image_sam_3d_body_mapper.py` itself — it has been restored to its original state.

The full regression test passed: https://github.com/datajuicer/data-juicer/actions/runs/28078847496


fengrui-z (commented):
The test skip for `image_sam_3d_body_mapper` looks good.

However, the model name update is incomplete. Beyond the 4 files changed here, there are still **170+ occurrences** of `qwen2.5-72b-instruct` scattered across:

- **17 operator doc files** under `docs/operators/` (mapper, filter, aggregator) — these serve as user-facing guidance and examples
- **30+ test files** under `tests/ops/`
- **`docs/op_doc_enhance_workflow/examples.json`** — also contains `qwen2.5-vl-3b-instruct`

Leaving these unchanged means users following the docs will hit deprecated model errors, and the codebase remains inconsistent with some references updated and others not.

**Suggestions:**

1. Update all remaining `qwen2.5-72b-instruct` → `qwen3.7-max` and `qwen2.5-vl-3b-instruct` → `qwen3-vl-8b-instruct` across docs and tests in this PR
2. If a full sweep is too large, consider introducing a centralized constant for default model names to avoid this kind of scattered update in the future
3. At minimum, the docs should be updated in this same PR since they directly affect user experience

cmgzn:
> * **`docs/op_doc_enhance_workflow/examples.json`**

Hi @fengrui-z, thanks for the detailed review.

1. **Regarding the comprehensive model name update**: I agree that the current PR isn't thorough enough. However, `docs/op_doc_enhance_workflow/examples.json` is a cache file automatically generated by `generate_op_details.py`, which should have been updated automatically when updating the op docs. That said, I believe the `op_doc_enhance_workflow` is outdated and effectively deprecated, so I'd prefer not to use it for updating the op docs. I plan to overhaul the entire workflow (including removing this file) in a follow-up PR, so I've left it untouched for now. I'll manually update the model names in `docs/operators/*.md` and the test files within this PR.

2. **Regarding introducing centralized constants**: Great suggestion! I'll push a commit shortly to add a unified default model constant.

fengrui-z (approved):
LGTM

Issue #982, opened 2026-05-20
data-juicer 有支持arm cpu架构吗？


qifengz:
### Before Submitting

- [x] I have asked the [Data-Juicer Q&A Copilot](https://datajuicer.github.io/data-juicer/en/main/docs_index.html) (available on Doc Site, DingTalk, or Discord) about this feature, but still want to submit it as a formal request.

- [x] I have searched the Data-Juicer [issues](https://github.com/datajuicer/data-juicer/issues) and found no similar feature requests.

### Description

目前arm架构下编译报错，如下部分日志：
[...]

PR #998, opened 2026-06-30
fix: unblock ARM64 install by marking decord/torchcodec with platform markers

cmgzn:
## Background

Related issue: #982

Data-Juicer currently cannot install `[vision]`, `[audio]`, or `[all]` on Linux ARM64 (e.g. `ubuntu-24.04-arm`), and also has gaps on macOS x86_64, because two dependencies lack wheels for those platforms:

- `decord`: only `manylinux2010_x86_64` / `win_amd64` wheels; the macOS x86_64 wheels only cover cp36–cp38 (Data-Juicer requires `>=3.10`)
- `torchcodec==0.7`: only `manylinux_2_28_x86_64` / `macosx_11_0_arm64` / `win_amd64` wheels; no Linux aarch64 or macOS x86_64 wheel

This blocks users on ARM64 Linux and macOS x86_64 from installing any extra that transitively includes them.

## Change

Restrict both packages to platforms that actually ship pre-built wheels, instead of only excluding `aarch64`:

```toml
"decord; (sys_platform == 'linux' and platform_machine == 'x86_64') or (sys_platform == 'win32' and platform_machine == 'AMD64')"
"torchcodec==0.7; (sys_platform == 'linux' and platform_machine == 'x86_64') or (sys_platform == 'darwin' and platform_machine == 'arm64') or (sys_platform == 'win32' and platform_machine == 'AMD64')"
```

Platform coverage after the change:

| package | Linux x86_64 | Linux aarch64 | macOS x86_64 | macOS arm64 | Windows x86_64 |
|---|---|---|---|---|---|
| `decord` | ✅ | ❌ | ❌ | ❌ | ✅ |
| `torchcodec==0.7` | ✅ | ❌ | ❌ | ✅ | ✅ |

- `decord` is an optional video backend in `video_utils.py` (`ffmpeg`/`av` are available as fallbacks).
- `torchcodec` is only used by two audio filters via `LazyLoader`.
- x86_64 Linux/Windows behavior is unchanged; macOS arm64 `torchcodec` is preserved (0.7 has a wheel).

## ARM64 CI verification

Tested on GitHub Actions `ubuntu-24.04-arm` (native `aarch64`):

| extra | before | after |
|---|---|---|
| core | ✅ | ✅ |
| `[nlp]` | ✅ | ✅ |
| `[distributed]` | ✅ | ✅ |
| `[generic]` | ✅ | ✅ |
| `[audio]` | ❌ torchcodec | ✅ |
| `[vision]` | ❌ decord | ✅ |
| `[all]` | ❌ | ✅ |

Notably, `[generic]` already installs fine on ARM64 (`vllm==0.11.0`, `cudf-cu12==25.10.0` have aarch64 wheels). The only blockers were `decord` and `torchcodec==0.7`.

## Notes

- `torchcodec` cannot be upgraded independently because 0.8+ requires `torch>=2.9`; Data-Juicer currently pins `torch==2.8.0`.
- On platforms without wheels, `decord`/`torchcodec` dependent operators are unavailable, but installation and other operators work.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ vision = [
     "scenedetect[opencv]",  # Scene detection
     "ultralytics",  # YOLO models
     "rembg",  # Background removal
-    "decord", # video/audio handling
+    "decord; (sys_platform == 'linux' and platform_machine == 'x86_64') or (sys_platform == 'win32' and platform_machine == 'AMD64')", # video/audio handling (no aarch64/macOS py3.10+ wheel)
     "qwen-vl-utils==0.0.14",  # for Qwen-VL
     "ptlflow==0.4.1",  # optical flow models collection
     "timm==1.0.22",  # avoid importing issue in the latest v1.0.23
@@ -104,7 +104,7 @@ audio = [
     "resampy",
     "samplerate==0.1.0",
     "torchaudio",  # Audio processing with PyTorch
-    "torchcodec==0.7",
+    "torchcodec==0.7; (sys_platform == 'linux' and platform_machine == 'x86_64') or (sys_platform == 'darwin' and platform_machine == 'arm64') or (sys_platform == 'win32' and platform_machine == 'AMD64')", # only platforms with 0.7 wheels
     "soundfile",  # audio handling
     "ffmpeg-python",
     "audiomentations",
--- a/uv.lock
+++ b/uv.lock
@@ -1910,7 +1910,7 @@ name = "decord"
 version = "0.6.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "numpy" },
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -71,7 +71,7 @@ vision = [
     "scenedetect[opencv]",  # Scene detection
     "ultralytics",  # YOLO models
     "rembg",  # Background removal
-    "decord", # video/audio handling
+    "decord; (sys_platform == 'linux' and platform_machine == 'x86_64') or (sys_platform == 'win32' and platform_machine == 'AMD64')", # video/audio handling (no aarch64/macOS py3.10+ wheel)
```
fengrui-z, line 74: Please regenerate uv.lock after changing dependency markers.

fengrui-z (changes requested):

pyproject.toml
cmgzn: [bca9f3d](https://github.com/datajuicer/data-juicer/pull/998/commits/bca9f3da256cc239906e78179a17b8af921df0b4)

fengrui-z (approved):
LGTM

PR #1001, opened 2026-06-30
fix(mapper): make clean_html_mapper robust to null text values

cmgzn:
Prevent CleanHtmlMapper from crashing when the configured text column contains None by returning an empty string for null values.

- Add a minimal None guard in CleanHtmlMapper.process_batched.
- Add a focused regression test for None text input.
- Verified: uv run python -m unittest tests.ops.mapper.test_clean_html_mapper (8/8 passed).

### DIFF
--- a/data_juicer/ops/mapper/clean_html_mapper.py
+++ b/data_juicer/ops/mapper/clean_html_mapper.py
@@ -34,7 +34,7 @@ class CleanHtmlMapper(Mapper):
 
     def process_batched(self, samples):
         def _clean_html(raw_html):
-            if raw_html is None:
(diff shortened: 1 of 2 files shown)


fengrui-z (commented):
1. Returning `""` silently swallows bad data. Consider also logging a warning so users know they have dirty inputs — this is what `extract_support_text_mapper.py:116` does with `logger.warning("Invalid input summary!")`.
2. The two test methods (`test_null_text` and `test_non_string_text`) could be merged into one parameterized test since they test the same behavior.

fengrui-z (approved):
LGTM

PR #1002, opened 2026-06-30
Add comprehensive tests for utility functions and model handling

cmgzn:
### Summary

Add comprehensive unit tests for utility modules, operator parse/output logic, and edge cases across the test suite. Also includes two minor bug fixes discovered during testing.

### Test Coverage Added

-   **Utility modules**: `common_utils`, `file_utils`, `compress`, `cache_utils`, `fingerprint_utils`, `nltk_utils`, `video_utils`, `webdataset_utils`, `lazy_loader`, `model_utils`, `llm_semantic_ops`
-   **Operator tests**: Parse-output and edge-case tests for LLM-based mappers/filters/aggregators (entity extraction, dialog normalization, sentiment analysis, keyword extraction, QA generation, PII redaction, etc.)
-   **Core modules**: `load_strategy`, `event_logging_mixin`, `exporter`, `formatter`, `op_fusion`
-   **Download module**: Wikipedia download tests with mocked HTTP responses

### Bug Fixes

-   **`common_utils.py`**: Fix `stats_to_number` condition check — `s == []` raises a `FutureWarning` with NumPy arrays; replaced with explicit `isinstance(s, list)` and `len(s) == 0`.
-   **`video_utils.py`**: Fix `FFmpegStreamingWriter.is_available()` — catch `OSError` / `subprocess.CalledProcessError` instead of `ImportError` (which `subprocess.run` never raises).


tests/utils/test_nltk_utils.py
```diff
@@ -0,0 +1,150 @@
...
+            self._with_nltk_path([nltk_data_dir], reset)
+            self.assertTrue(os.path.isdir(nltk_data_dir))
+            self.assertFalse(os.path.exists(keep_file))
+
+            def empty(_nltk):
+                nltk_utils.clean_nltk_cache()
+
+            self._with_nltk_path([], empty)
+
+    def test_patch_nltk_pickle_security_replaces_loader_and_allows_classes(self):
```
fengrui-z: In `tests/utils/test_nltk_utils.py:76-93`, the new test requires `restricted_pickle_load` to accept arbitrary `pickle.dumps(...)` payloads.

The tested implementation replaces NLTK’s restricted loader with direct calls to `pickle.load`. Malicious pickle payloads can execute arbitrary code through reducers. Although the runtime behavior predates this PR, this test turns it into a protected contract and would make a future security fix fail.

Please keep the restricted loader and extend only the required class allowlist. The test should also verify that a malicious reducer payload is rejected.

fengrui-z:
The pickle security concern remains unresolved: patch_nltk_pickle_security() still replaces the restricted loader with pickle.load, while the updated test does not verify that malicious reducer payloads are rejected.
Please preserve the restricted loader, extend only the required allowlist, and add a negative security test. The PR description should also be updated to reflect its significantly expanded scope.

fengrui-z (approved):

PR #1004, opened 2026-07-09
feat: add FusedSequentialBatchOp for batch-local stage fusion

fengrui-z:
## Summary

This PR adds `FusedSequentialBatchOp`, a batch-local fused operator for reducing Ray/Data stage overhead.

The original direction in this PR explored fusing multiple GPU mappers through a `FusedParallelMapper`-style execution model. During review and validation, that scope turned out to mix several different fusion problems:

1. preserving normal pipeline order across mapper/filter ops
2. sharing decoded/intermediate context such as words, lines, images, videos, and audio
3. running read-only analyzers/taggers in parallel when they only write disjoint meta/stats

Those have different safety contracts, so this PR is narrowed to the first and safest piece: sequential batch-local fusion.

`FusedSequentialBatchOp` runs compatible sub-ops in the original order inside one batch stage. It preserves normal dependency semantics, passes each returned batch to the next sub-op, and supports filters that compute stats and drop rows.

## Problem

Ray/Data pipelines pay overhead for every `map_batches` / `filter` stage:

- each op becomes a separate dataset stage
- batches are materialized and scheduled between stages
- model-backed ops may have repeated actor/stage overhead
- chains of small or medium batch-local ops can spend non-trivial time in orchestration rather than useful work

The initial `FusedParallelMapper` direction attempted to reduce this by running independent mappers together inside one stage. However, that introduced a stronger and riskier contract:

- sub-ops had to be independent
- sub-ops had to be read-only except for disjoint writes
- returned batches could not safely be chained
- filters/drop-row semantics were not naturally represented
- shared decode/context reuse was a separate concern
- thread/CUDA stream parallelism belonged to a different optimization class

Because of that, the PR scope was cleaned up to avoid mixing multiple fusion strategies in one change.

## Solution

This PR introduces `FusedSequentialBatchOp`.

It focuses only on stage reduction while preserving the original pipeline semantics:

- sub-ops run sequentially in config order
- each sub-op receives the batch returned by the previous sub-op
- mapper outputs can be consumed by later ops in the same fused stage
- filter sub-ops can compute stats and drop rows
- metadata/stat columns are initialized when needed
- Ray scheduling kwargs are stripped before constructing inner sub-ops
- profiling records per-sub-op wall time inside the fused stage

This makes the fused op behave like a local mini-pipeline inside one Ray/Data batch stage, rather than a parallel fan-out/fan-in executor.

### Added `FusedSequentialBatchOp`

`FusedSequentialBatchOp` supports two construction modes:

- `op_specs`: build sub-ops lazily from registry names and kwargs
- `fused_ops`: reuse pre-built op instances from auto-fusion

It currently supports batch-local mapper/filter ops.

For mapper-like sub-ops:

- calls `op.process(...)`
- validates that a dict batch is returned
- passes the returned batch to the next sub-op

For filter sub-ops:

- initializes `Fields.stats` when needed
- calls `op.compute_stats(...)`
- calls `op.process(...)` to get the keep mask
- applies the keep mask to every batch column
- passes the filtered batch to the next sub-op

### Updated mapper fusion integration

The existing mapper fusion path now creates `FusedSequentialBatchOp`.

Auto-fusion remains conservative:

- only GPU `Mapper` ops are considered
- each op must explicitly opt in with `_fused_sequential_batch_op_safe = True`
- ops must declare disjoint output columns
- ops must not read columns produced earlier in the same group
- aggregate GPU budget must stay under the configured limit

This preserves the safety checks while changing the fused executor from the previous parallel direction to sequential batch-stage reduction.

### Scope cleanup

This PR removes the old parallel-mapper surface from the final code:

- removed `FusedParallelMapper`
- removed `fused_parallel_mapper` registry name
- removed `_fused_parallel_mapper_safe`
- removed thread/stream-based parallel execution behavior from this PR

The only fused op introduced by this PR is `FusedSequentialBatchOp`.

## Out of Scope

This PR intentionally does not implement the other fusion classes discussed during review.

### `FusedSharedContextOp`

Not included here.

This should separately handle cases where multiple ops can reuse shared context, such as:

- words
- lines
- image decode
- video decode / sampled frames
- audio decode
- other intermediate context

That requires a separate contract for context lifecycle, cache ownership, and cleanup.

### `FusedParallelReadOnlyOp`

Not included here.

This should separately handle explicitly opted-in read-only analyzer/tagger ops that can run in parallel and only write disjoint meta/stats.

That requires a stricter contract:

- no row dropping
- no top-level mutation except declared outputs
- no dependencies between sub-ops
- disjoint meta/stats writes
- explicit read-only opt-in

Keeping this separate avoids mixing sequential pipeline semantics with parallel fan-out semantics.

## Tests

Added/updated tests for:

- `FusedSequentialBatchOp` sequential mapper execution
- chaining returned batches between sub-ops
- filter sub-ops computing stats and dropping rows
- `op_specs` construction from registry
- stripping Ray scheduling kwargs from inner sub-op kwargs
- cleanup columns
- empty batches
- mapper fusion opt-in checks
- dependency/write-overlap checks
- config-level `mapper_fusion` behavior

Validated locally with:

```bash
pytest -q tests/ops/test_mapper_fusion.py
pytest -q tests/config/test_config.py::ConfigTest::test_op_fusion
python3 -m compileall -q data_juicer/ops/fused_sequential_batch_op.py data_juicer/ops/op_fusion.py data_juicer/ops/__init__.py tests/ops/test_mapper_fusion.py data_juicer/config/config.py
git diff --check


Qirui-jiao (approved):

fengrui-z:
### Reuse of existing `GeneralFusedOP`

`GeneralFusedOP` and `FusedSequentialBatchOp` now share a common
batch-local execution engine in `fused_batch_executor.py`.

The shared engine owns:

- sequential Mapper/Filter execution
- chaining returned batches
- filter mask application
- optional meta/stats initialization
- optional shared-context lifecycle
- optional strict result validation
- per-op timing callbacks

The two public operators retain different execution policies:

- `GeneralFusedOP` preserves explicit configuration, input deepcopy,
  shared context, `LazyBatch` compatibility, non-batched fallback, and
  dataset-level preparation.
- `FusedSequentialBatchOp` preserves lazy worker initialization, standard
  OP wrappers, strict batch/mask validation, profiling, and Ray resource
  planning for automatic GPU stage fusion.

This avoids maintaining two independent sequential execution loops while
preserving the existing `GeneralFusedOP` behavior.

cmgzn:
LGTM. This PR is self-contained fusion infrastructure, and in practice it won't actually be activated after merging — no operator currently declares `_fused_sequential_batch_op_safe` / `estimated_vram_fraction`, so it has zero impact on existing pipelines and is safe.

For follow-up, I have a minor concern about relying on `_fused_sequential_batch_op_safe` and `estimated_vram_fraction` — two manually-authored attributes — as the switches for automatic fusion. I'd like to leave these as discussion points for later:

1. These attributes are currently based purely on manual declaration, and a wrong value won't raise any error. In the future we might want to add some tests to verify that operators declaring these attributes can indeed be fused correctly.
2. `estimated_vram_fraction` is a static estimate, whereas actual VRAM usage varies with runtime factors such as batch size and input resolution/length. Using a static value for the VRAM budget check may be worth reconsidering in terms of reliability.

Separately, it might be worth adding some fusion-related documentation down the line, clarifying the difference between manual and automatic fusion, as well as the meaning of these config options / attributes.

None of this blocks the merge in my view. Thanks for the careful work here!

PR #1016, opened 2026-07-24
feat: add RayAnalyzer for distributed data analysis

### LINKED ISSUE #1008: Why doesn’t the analyzer support Ray mode?
### Before Asking

- [x] I have asked the [Data-Juicer Q&A Copilot](https://datajuicer.github.io/data-juicer/en/main/docs_index.html) (available on Doc Site, DingTalk, or Discord), but the problem still persists.

- [x] I have read the [README](https://github.com/datajuicer/data-juicer/blob/main/README.md) carefully.

- [x] I have pulled the latest code of main branch to run again and the problem still existed.

### Search before asking

- [x] I have searched the Data-Juicer [issues](https://github.com/datajuicer/data-juicer/issues) and found no similar questions.

### Question

Why doesn’t the analyzer support Ray mode? Wouldn’t processing millions of data samples in the sandbox be extremely slow?

### Environment Information

_No response_

### Additional Context

_No response_

cmgzn:
## What

Add `RayAnalyzer` for distributed data analysis using Ray, as the counterpart to the local `Analyzer`.

## Changes

- **`data_juicer/core/ray_analyzer.py`** (new): RayAnalyzer class that computes filter stats via Ray `map_batches` and aggregates overall statistics (count/mean/std/min/max) using Ray native aggregation operators — no pandas materialization
- **`data_juicer/core/data/ray_dataset.py`**: Add `stats_only` parameter to `process()` and `_run_single_op()` so stats can be computed without actually filtering rows
- **`tools/analyze_data.py`**: Dispatch `RayAnalyzer` when `executor_type` is `ray` or `ray_partitioned`
- **`data_juicer/config/config.py`** / **`data_juicer/core/__init__.py`**: Register RayAnalyzer
- **`docs/`**: Add Distributed Data Analysis sections to Distributed.md/Distributed_ZH.md, update QuickStart docs
- **`demos/process_on_ray/configs/analyze.yaml`**: Demo config for Ray analysis

## Key Design Decisions

- Uses Ray native `Mean`/`Std`/`Min`/`Max` aggregation — avoids materializing full dataset into pandas
- No per-column distribution charts, correlation analysis, or percentiles in Ray mode — focuses on scalable overall statistics
- Same `dj-analyze` CLI entry point, automatically dispatched by `executor_type`
- Falls back to local `ray.init()` when no existing Ray cluster is found

## Testing

- Verified with 1M-row dataset: stats computation 15s, aggregation 1.9s
- Verified full sandbox pipeline (Probe → Refine → Execute → Evaluate) with Ray mode
- Demo config tested end-to-end

### DIFF
--- a/data_juicer/core/ray_analyzer.py
+++ b/data_juicer/core/ray_analyzer.py
@@ -57,7 +57,11 @@ class RayAnalyzer:
 
         from data_juicer.utils.ray_utils import initialize_ray
 
-        initialize_ray(cfg=cfg)
+        try:
+            initialize_ray(cfg=self.cfg)
+        except ConnectionError:
+            logger.info("No existing Ray cluster found, starting a local one...")
+            ray.init(ignore_reinit_error=True)
 
         self.dataset_builder = DatasetBuilder(self.cfg, executor_type="ray")
 
--- a/data_juicer/core/ray_analyzer.py
+++ b/data_juicer/core/ray_analyzer.py
@@ -3,6 +3,7 @@ from typing import Optional
 
 import pandas as pd
 import pyarrow
+import pyarrow.compute as pc
 from jsonargparse import Namespace
 from loguru import logger
 from pydantic import PositiveInt
@@ -52,7 +53,7 @@ class RayAnalyzer:
     """
 
     def __init__(self, cfg: Optional[Namespace] = None):
-        self.cfg = init_configs(which_entry=self) if cfg is None else cfg
+        self.cfg = init_configs(allow_auto=True) if cfg is None else cfg
         self.work_dir = self.cfg.work_dir
 
         from data_juicer.utils.ray_utils import initialize_ray
@@ -142,7 +143,7 @@ class RayAnalyzer:
     def _compute_overall(self, dataset):
         os.makedirs(self.analysis_path, exist_ok=True)
 
-        from ray.data.aggregate import Max, Mean, Min, Std
+        from ray.data.aggregate import Count, Max, Mean, Min, Std
 
         select_cols = []
         available_cols = dataset.data.columns()
@@ -163,6 +164,8 @@ class RayAnalyzer:
             return pd.DataFrame()
 
         numeric_cols = []
+        list_numeric_cols = []
+        skipped_cols = []
         arrow_schema = flat_data.schema()
         if hasattr(arrow_schema, "base_schema"):
             arrow_schema = arrow_schema.base_schema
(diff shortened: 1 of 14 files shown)


tools/analyze_data.py
```diff
@@ -1,11 +1,38 @@
...
+    for i, arg in enumerate(args):
+        if arg == "--executor_type" and i + 1 < len(args):
+            return args[i + 1]
+    for i, arg in enumerate(args):
+        if arg == "--config" and i + 1 < len(args):
+            try:
+                import yaml
+
+                with open(args[i + 1]) as f:
+                    cfg = yaml.safe_load(f)
```
fengrui-z: This pre-parser only recognizes --executor_type ray and --config path. However, jsonargparse also accepts --executor_type=ray and --config=path. With either form, _get_executor_type() returns "default" and silently instantiates the local Analyzer, which could cause a large Ray workload to be loaded and analyzed locally.
Could we reuse the normal configuration parser to determine executor_type, or at least handle both the separated and --key=value forms here?

data_juicer/core/ray_analyzer.py
```diff
@@ -0,0 +1,201 @@
...
+
+        numeric_cols = []
+        arrow_schema = flat_data.schema()
+        if hasattr(arrow_schema, "base_schema"):
+            arrow_schema = arrow_schema.base_schema
+        for col_name in flat_cols:
+            idx = arrow_schema.get_field_index(col_name)
+            field_type = arrow_schema.field(idx).type
+            if pyarrow.types.is_integer(field_type) or pyarrow.types.is_floating(field_type):
+                numeric_cols.append(col_name)
```
fengrui-z, line 176: This type check only accepts scalar integer and floating-point Arrow columns. Many existing multimodal filters produce list-valued numeric statistics, such as image sizes, frame scores, and video/audio durations. Those columns will therefore be silently omitted from overall.csv.
The local OverallAnalysis explodes list-valued columns before calculating their descriptive statistics. Could the Ray implementation similarly flatten numeric list columns before aggregation, or explicitly report unsupported columns instead of silently dropping them?

data_juicer/core/ray_analyzer.py
```diff
@@ -0,0 +1,201 @@
...
+        aggs = []
+        for col in numeric_cols:
+            aggs.extend([Mean(col), Std(col), Min(col), Max(col)])
+
+        agg_result = flat_data.aggregate(*aggs)
+
+        rows = {}
+        for col in numeric_cols:
+            rows[col] = {
+                "count": total_count,
```
fengrui-z: total_count is the number of dataset rows, but Ray’s Mean, Std, Min, and Max aggregations ignore nulls by default. Some filters can legitimately produce None, so the reported count can differ from the number of values used to calculate the other statistics.
Could we add Count(col) for each numeric column and use that result here? Including the counts in the same aggregate() call would also remove the separate flat_data.count() action.

cmgzn: [3668755](https://github.com/datajuicer/data-juicer/pull/1016/commits/3668755eceed26d2d21e6f44b7c8d65f13a4afce)

data_juicer/core/ray_analyzer.py
cmgzn: [3668755](https://github.com/datajuicer/data-juicer/pull/1016/commits/3668755eceed26d2d21e6f44b7c8d65f13a4afce)

tools/analyze_data.py
cmgzn: [3668755](https://github.com/datajuicer/data-juicer/pull/1016/commits/3668755eceed26d2d21e6f44b7c8d65f13a4afce)

yxdyc (approved):

PR #1020, opened 2026-07-28
fix: misc bug fixes and bump fsspec for pyarrow>=17 compat

cmgzn:
## Summary

A collection of small bug fixes and a dependency bump.

### Changes

1. **fix(image_diffusion_mapper)**: handle missing or null captions gracefully — prevents crash when samples have missing or empty caption fields. (Related: cmgzn/data-juicer#60)

2. **fix: warn when process operator list is empty** — adds a warning in both `DefaultExecutor` and `RayExecutor` when the user provides an empty process operator list, preventing silent no-op execution. (Related: cmgzn/data-juicer#69)

3. **fix(vggt_mapper)**: guard and initialize `__dj__meta__` before access — ensures `__dj__meta__` is initialized before `VggtMapper` accesses it, preventing `KeyError` on samples without prior metadata. (Related: cmgzn/data-juicer#93)

4. **deps: bump fsspec from `==2023.5.0` to `>=2026.1.0`** — the old pin is incompatible with `pyarrow>=17`. Root cause: `fsspec<=2025.9.0` mirrors `ArrowFile.size` from the pyarrow stream via `@mirror_from` (treats it as an attribute), but `pyarrow>=17` changed `NativeFile.size` to a method. This causes `TypeError: unsupported operand type(s) for +: 'method' and 'float'` in tqdm during HDFS downloads through `ArrowFSWrapper`. `fsspec>=2026.1.0` replaces the mirror with an explicit `@property` that calls `stream.size()`, returning the correct int. Verified end-to-end with `fsspec==2026.4.0` + `pyarrow==24.0.0` on a Docker HDFS setup (load → filter → export).

   Note: `datasets==5.0.0` declares `fsspec<=2026.4.0`, so the effective resolved range is `2026.1.0–2026.4.0`. `s3fs` versions are co-released with matching fsspec versions (e.g. `s3fs==2026.4.0` requires `fsspec==2026.4.0`), so no resolver conflicts.

### DIFF
--- a/data_juicer/ops/mapper/image_diffusion_mapper.py
+++ b/data_juicer/ops/mapper/image_diffusion_mapper.py
@@ -155,13 +155,24 @@ class ImageDiffusionMapper(Mapper):
 
         # load captions
         if self.caption_key:
-            captions = ori_sample[self.caption_key]
-            if not isinstance(captions, list):
+            captions = ori_sample.get(self.caption_key)
+            num_images = len(loaded_image_keys)
+            if not captions:
+                # no caption at all (missing key, None or an empty list): an
+                # empty prompt is a valid unconditional prompt for
+                # image-to-image generation, so fall back to it
+                captions = [""] * num_images
+            elif not isinstance(captions, list):
                 # one caption for all images
-                captions = [captions] * len(images)
+                captions = [captions] * num_images
             else:
-                assert len(captions) == len(images), "The num of captions must match the num of images."
-            captions = [remove_special_tokens(c) for c in captions]
+                # a partially filled list carries no hint about which image the
+                # missing captions belong to, so keep failing loudly here; use
+                # None or "" as an explicit placeholder to skip a single image
+                assert len(captions) == num_images, "The num of captions must match the num of images."
+            # a blank entry inside the list is an explicit "no guidance for this
+            # image" placeholder, so normalize it instead of skipping
+            captions = [remove_special_tokens(c) if isinstance(c, str) else "" for c in captions]
         else:
             caption_samples = {
                 self.text_key: [SpecialTokens.image] * len(images),
@@ -226,6 +237,10 @@ class ImageDiffusionMapper(Mapper):
                 samples_after_generation.extend(generated_samples)
 
         # reconstruct samples from "list of dicts" to "dict of lists"
+        if not samples_after_generation:
+            # every sample in this batch was skipped (e.g. no images at all);
+            # keep the "dict of lists" schema so downstream stays consistent
+            return {key: [] for key in samples}
         keys = samples_after_generation[0].keys()
         res_samples = {}
         for key in keys:
--- a/data_juicer/core/executor/default_executor.py
+++ b/data_juicer/core/executor/default_executor.py
@@ -167,6 +167,11 @@ class DefaultExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin):
         logger.info("Preparing process operators...")
         ops = load_ops(self.cfg.process)
 
+        if not ops:
+            logger.warning(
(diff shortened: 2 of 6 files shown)


data_juicer/ops/mapper/image_diffusion_mapper.py
```diff
@@ -155,7 +156,12 @@ def _process_single_sample(self, ori_sample, rank=None, context=False):
 
         # load captions
         if self.caption_key:
-            captions = ori_sample[self.caption_key]
+            captions = ori_sample.get(self.caption_key)
+            if captions is None or (isinstance(captions, str) and not captions.strip()):
+                logger.warning(
+                    f"Sample has missing or empty caption for key " f"'{self.caption_key}', skipping generation."
+                )
+                return []
```
fengrui-z: Missing captions can still crash an all-invalid batch.** Returning `[]` from `_process_single_sample()` leaves `samples_after_generation` empty when `keep_original_sample=False`, after which `process_batched()` accesses `samples_after_generation[0]` and raises `IndexError`. Please handle an empty output batch explicitly and add regression tests for missing, `None`, blank-string, and empty-list captions.

cmgzn: Thanks @fengrui-z — you're right, and digging into it turned up a bit more than the `IndexError`.

**Confirmed the crash.** With `keep_original_sample=False` and every sample in a batch skipped, `samples_after_generation` is empty and `samples_after_generation[0].keys()` raises `IndexError`. Worth noting this line had no empty-batch guard before this PR either; the new `return []` path just made it much easier to hit.

**But the root cause was one layer deeper: empty captions shouldn't be skipped at all.** I tested the pre-PR code and an empty-string caption was already generating fine — `_real_guidance` gets called with `prompt=''`, which is a perfectly valid unconditional prompt for img2img. The operator also generates captions via BLIP2 when `caption_key` is unset, so "no caption" was never an error state by design.

So the original crashes weren't about empty prompts being unsupported — they were plain robustness gaps: `ori_sample[key]` → `KeyError`, `remove_special_tokens(None)` → `AttributeError`, `[]` → `AssertionError`. Skipping those samples silently dropped data and, ironically, turned off a path that already worked.

Reworked accordingly:

1. Empty captions (missing key / `None` / `[]`) now normalize to an empty prompt and generate as usual, instead of being skipped.
2. Added the empty-batch guard, returning `{key: [] for key in samples}` to keep the dict-of-lists schema. Still needed, since the no-images path is a genuine skip.
3. Fixed a pre-existing alignment bug: captions were aligned against `len(images)`, but `images` is a dict keyed by path, so duplicate paths collapse it and indexing overflows against `loaded_image_keys`. Now aligned on `len(loaded_image_keys)`. The old `assert` was masking this — it only surfaced once the length check was relaxed.
4. Kept the `assert` for *partially* filled caption lists. Padding those would silently pair captions with the wrong images, which is worse than failing loudly. Use `None` or `""` as an explicit per-image placeholder — position is preserved.

Added 11 regression tests covering missing / `None` / blank-string / empty-list captions across both `keep_original_sample` branches, plus placeholder alignment (including a leading `None`) and the duplicate-path case. All verified to fail before the fix and pass after. `_real_guidance` is stubbed so they run offline without downloading a diffusion model.

One judgement call worth flagging: I considered reusing the existing `op_generate_caption` to fill in missing captions, which would be more consistent with the `caption_key=None` path. I skipped it because `ImageCaptioningMapper` declares `memory="16GB"` and pulls a ~15GB BLIP2 checkpoint, and `op_generate_caption` is only constructed when `caption_key` is falsy — making it unconditional would raise this operator's resource footprint from 8GB for everyone, including users with no empty captions. Empty prompts are zero-cost and match the pre-existing behaviour.

yxdyc (approved):

PR #1014, opened 2026-07-22
feat: support HDFS protocol for dataset loading and exporting

LiuGuH:
<h2>PR Description</h2><h3>What does this PR do?</h3><p>This PR adds HDFS (<code>hdfs://</code>) protocol support to data-juicer, enabling users to read datasets from and write processed results to HDFS — consistent with the existing S3 remote storage workflow.</p><h3>Motivation</h3><p>Many production environments store large-scale datasets on HDFS clusters. Supporting HDFS natively allows users to process data in-place without manually staging files to local disk first.</p><h3>Changes</h3><ul><li><strong><code>data_juicer/utils/hdfs_utils.py</code></strong> (new): HDFS utility module — path validation (<code>validate_hdfs_path</code>, <code>strip_hdfs_scheme</code>) and PyArrow <code>HadoopFileSystem</code> creation with configurable host/port/user/Kerberos/extra_conf.</li><li><strong><code>data_juicer/core/data/load_strategy.py</code></strong>: Added <code>DefaultHdfsDataLoadStrategy</code> (local executor) and <code>RayHdfsDataLoadStrategy</code> (Ray executor), registered via <code>DataLoadStrategyRegistry</code> under <code>("default"/"ray", "remote", "hdfs")</code>. Uses <code>fsspec</code> <code>ArrowFSWrapper</code> for filesystem abstraction.</li><li><strong><code>data_juicer/core/exporter.py</code></strong> / <strong><code>ray_exporter.py</code></strong>: Extended exporters to detect <code>hdfs://</code> export paths and create a PyArrow HDFS filesystem for writing output.</li><li><strong><code>data_juicer/config/config.py</code></strong>: Generalized the S3-only remote path handling to cover both <code>s3://</code> and <code>hdfs://</code> schemes (log filename generation, <code>work_dir</code> fallback, etc.).</li><li><strong>Demos</strong>: Added <code>demos/process_simple/process_hdfs.yaml</code> (local executor) and <code>demos/process_on_ray/configs/demo_hdfs.yaml</code> (Ray executor) as usage examples.</li><li><strong>Tests</strong>: Comprehensive unit tests for HDFS load strategies, exporters, and <code>hdfs_utils</code>.</li></ul><h3>Usage</h3><p>Users can now specify HDFS paths in their YAML config:</p><pre><code>dataset:
  configs:
    - type: remote
      source: hdfs
      format: jsonl
      path: 'hdfs://namenode:8020/user/data/demo-dataset.jsonl'

[...]

PR #1021, opened 2026-07-29
Feat: Unify remote filesystem dispatch for exporters (groundwork for lakehouse support)

cmgzn:
# Motivation

This is the first groundwork PR in the incremental split of #911 (lakehouse data source support). It converges the duplicated "detect path prefix → extract credentials/config → build a PyArrow FileSystem" logic scattered across exporters into a single dispatch extension point. Follow-up Iceberg/Delta/Hudi PRs only need to extend the scheme branches in one place instead of re-implementing per-callsite logic.

Builds on the HDFS support introduced in #1014.

# Changes

- **New `data_juicer/utils/fs_utils.py`** with `create_filesystem_for_path(path, extra_args=None) -> (fs, remaining_args)`:
  - dispatches by URI scheme: `s3://` → `create_pyarrow_s3_filesystem`, `hdfs://` → `create_pyarrow_hdfs_filesystem` (config includes `path` so host/port can be inferred), anything else → `(None, copy)`;
  - **side-effect-free**: never mutates the input `extra_args` dict; the backend-specific keys (`S3_FS_KEYS` / `HDFS_FS_KEYS`) are consumed and removed from the returned `remaining_args` copy;
  - pyarrow-dependent imports stay lazy (function-level), keeping the module lightweight to import;
  - module docstring documents the convention: future backends (Iceberg/Delta/Hudi) should extend the scheme branches here rather than adding per-callsite prefix logic.
- **Refactor `RayExporter`**: the two parallel S3/HDFS `if` blocks that popped `aws_*` / `hdfs_*` keys from `export_extra_args` are replaced by a single call to the helper; `self.s3_filesystem` / `self.hdfs_filesystem` attributes and their usage in write methods are unchanged.
- **Refactor `Exporter`**: the HDFS branch (PyArrow fs + `ArrowFSWrapper`) now delegates to the helper; the S3 branch keeps building fsspec `storage_options` (different backend contract) but its entry check is aligned to the same scheme semantics.
- **Case-insensitive scheme handling**: `validate_s3_path` / `validate_hdfs_path` and the exporter branch checks now treat the scheme case-insensitively, consistent with `urlparse` semantics.

# Behavior notes

- Single-slash typos such as `s3:/bucket/x` or `hdfs:/user/x` now raise a clear `ValueError` at exporter construction time (previously they were silently treated as local paths).
- Uppercase scheme variants (`S3://...`, `HDFS://...`) are now correctly accepted and routed to the corresponding filesystem.

# Related

- References #911 — original lakehouse PR; this PR is the first step of its split plan.
- Builds on #1014 — HDFS protocol support.

# Tests

- New `tests/utils/test_fs_utils.py` (15 cases): local/relative/`file://` paths, S3/HDFS dispatch with mocked factories (exact conf/remaining-key assertions), input-dict immutability (deep-compare), uppercase schemes, query-parameter paths, malformed paths (`ValueError`), `extra_args=None`.
- Extended `tests/utils/test_s3_utils.py` / `test_hdfs_utils.py` with uppercase-accept and single-slash-reject cases.
- Locally green: `tests/utils/test_fs_utils.py` + `test_s3_utils.py` + `test_hdfs_utils.py` (49 passed), `tests/core/test_exporter.py` + `tests/core/test_ray_exporter.py` (32 passed), load-strategy HDFS/S3 tests unaffected.

### DIFF
--- a/data_juicer/core/exporter.py
+++ b/data_juicer/core/exporter.py
@@ -1,12 +1,14 @@
 import json
 import os
 from multiprocessing import Pool
+from urllib.parse import urlparse
 
 from datasets import Dataset as HFDataset
 from loguru import logger
 
 from data_juicer.utils.constant import Fields, HashKeys
 from data_juicer.utils.file_utils import Sizes, byte_size_to_size_str
+from data_juicer.utils.fs_utils import create_filesystem_for_path
 
 
 class Exporter:
@@ -88,7 +90,8 @@ class Exporter:
 
         # Check if export_path is S3 and create storage_options if needed
         self.storage_options = None
-        if export_path.startswith("s3://"):
+        export_scheme = urlparse(export_path).scheme.lower()
+        if export_scheme == "s3":
             # Extract AWS credentials from kwargs (if provided)
             s3_config = {}
             if "aws_access_key_id" in kwargs:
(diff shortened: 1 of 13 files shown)


data_juicer/core/exporter.py
fengrui-z: Encryption skip — won't skip for S3:// paths

data_juicer/core/ray_exporter.py
fengrui-z: Encryption skip — same issue

data_juicer/core/exporter.py
fengrui-z: Shard path construction — falls through to local-path branch for `S3://` /`HDFS://`

cmgzn: 025780274c

cmgzn: 025780274c

data_juicer/core/ray_exporter.py
cmgzn: 025780274c

fengrui-z (approved):
LGTM

Dludora (approved):

PR #1022, opened 2026-07-29
[Bugfix] Process partitioned Ray datasets concurrently

Qirui-jiao:
## Summary

This PR fixes serial partition execution and hardens resource planning in
`PartitionedRayExecutor`.

The executor previously split a Ray Dataset into multiple partitions but
processed them one at a time on the driver. Increasing
`partition.num_of_partitions` therefore added independent Ray Data job and
Actor startup overhead without creating partition-level concurrency.

This change:

- drives partition jobs concurrently with a driver-side `ThreadPoolExecutor`;
- keeps the actual data processing and resource scheduling in Ray;
- preserves the original partition order when the processed datasets are
  unioned;
- reuses the existing checkpointing and event-logging path for every
  partition;
- calculates automatic operator parallelism once as a global job plan and
  scales it across concurrent partitions;
- treats an explicit Actor `num_proc` as a global pool budget instead of
  repeating the full pool in every partition;
- changes the default `partition.max_concurrent_partitions` from `32` to
  resource-aware `"auto"`;
- adds regression coverage for execution overlap, ordering, resource scaling,
  explicit Actor budgets, automatic partition limits, and restoration after
  success or failure.

No operator API or output schema is changed. Existing positive integer values
for `partition.max_concurrent_partitions` remain valid and override automatic
detection.

## Motivation

Partitioning is expected to let independent dataset partitions make progress
at the same time when Ray has sufficient resources. The previous implementation
used the following effective execution order:

```text
partition 0: submit -> materialize -> finish
partition 1: submit -> materialize -> finish
partition 2: submit -> materialize -> finish
...
```

`_process_with_checkpointing()` materializes the processed Ray Dataset before
returning. Because `_process_with_simple_partitioning()` called it from a
regular Python `for` loop, the driver waited for each partition to finish
before submitting the next one.

As a result:

- partition-level concurrency was always 1;
- idle CPUs or GPUs could not be used by another partition;
- GPU Actors were repeatedly created and destroyed in series;
- a larger partition count could make the job substantially slower.

This behavior was especially visible with one-GPU model Actors. On a four-GPU
node, `partition=4` still processed every row on GPU 0 while the other three
GPUs remained unused.

## Root cause

Ray Data transformations are lazy, but every partition is explicitly
materialized inside the existing checkpointing path:

```python
processed_dataset = dataset.process(ops)
processed_dataset.data = processed_dataset.data.materialize()
```

The outer serial loop therefore became a synchronization boundary around each
partition. Ray could parallelize blocks within one partition, but it never saw
multiple independent partition jobs at the same time.

The issue is in Data-Juicer's driver-side orchestration rather than in Ray's
partition or scheduling implementation.

Enabling concurrent partition drivers also exposed two resource-planning
risks:

1. An automatic or explicit Actor pool size could be reused independently by
   every partition. For example, four concurrent partitions could each see
   `num_proc=4` and collectively request up to 16 Actors even though the
   non-partitioned meaning is a four-Actor job budget.
2. The fixed default `max_concurrent_partitions=32` could start many Ray Data
   execution plans, driver threads, model initializations, and checkpoint
   writes at once, even on a four-GPU node.

The final implementation therefore separates global operator planning from
per-partition execution and resolves a safe outer-pipeline limit before
starting driver threads.

### Concurrent partition orchestration

Replace the serial partition loop with a bounded `ThreadPoolExecutor`:

```python
requested_max_workers = min(
    len(partitions),
    self.max_concurrent_partitions,
)
max_workers = self._limit_partition_workers_for_explicit_actors(
    ops,
    requested_max_workers,
)

with ThreadPoolExecutor(
    max_workers=max_workers,
    thread_name_prefix="ray-partition",
) as executor:
    futures = {
        executor.submit(
            self._process_partition,
            partition,
            partition_id,
            len(partitions),
            ops,
        ): partition_id
        for partition_id, partition in enumerate(partitions)
    }
```

The Python threads only drive independent Ray Dataset executions and wait for
their results. CPU/GPU data processing still runs in Ray tasks and Actors, so
Ray remains responsible for resource admission and placement.

### Plan operator resources globally

Operator resource planning now runs once on the driver before any partition
thread starts:

```python
self._configure_operator_parallelism(ops)
self._resolve_max_concurrent_partitions(ops)
```

This avoids concurrent mutation of the shared operator objects and preserves
the non-partitioned meaning of operator concurrency.

For automatically planned operators, the global concurrency is divided across
the number of concurrently driven partitions. Automatic values use ceiling
division so every partition retains usable concurrency. For example, on a
four-GPU node:

```text
global GPU num_proc=4

P=1 -> 1 partition worker  × num_proc=4
P=2 -> 2 partition workers × num_proc=2
P=4 -> 4 partition workers × num_proc=1
```

The original global values are restored in a `finally` block for later
convergence stages and for correct cleanup after partition failures.

### Treat explicit Actor num_proc as a global budget

A positive explicit Actor `num_proc` (including supported tuple/list forms) is
recorded as a global job budget. The executor:

- limits concurrent partition workers to the smallest explicit Actor budget;
- uses floor division to derive a safe per-partition Actor pool size;
- restores the configured value after partition processing.

For example, `num_proc=4` with four concurrent partitions becomes one Actor
per partition rather than four Actors per partition. If `num_proc=2` and four
partitions are requested, at most two partitions run concurrently, each with
one Actor.

Explicit Ray task concurrency is intentionally unchanged; this budget
handling applies to Actor pools whose instances and model memory would
otherwise be multiplied per partition.

### Resolve partition concurrency from Ray resources

`partition.max_concurrent_partitions` now defaults to `"auto"` and is resolved
after operator resource planning:

- GPU pipelines use the tightest CPU/GPU worker capacity reported by the Ray
  cluster;
- CPU-only pipelines use a conservative outer-pipeline cap of 4 because Ray
  Data already parallelizes blocks inside each partition;
- the result is bounded by the actual partition count and explicit Actor
  budgets;
- inability to inspect usable Ray resources falls back to one partition
  pipeline;
- a positive integer remains available as an explicit override.

This controls driver threads and the number of simultaneous Ray Data execution
plans. Ray still performs final CPU, GPU, memory, and object-store admission.

### Preserve deterministic union order

Concurrent partitions can finish in any order. The implementation therefore
preallocates the result list and stores every result by `partition_id`:

```python
processed_partitions = [None] * len(partitions)

for future in as_completed(futures):
    partition_id, processed_data = future.result()
    processed_partitions[partition_id] = processed_data
```

The existing union loop then consumes this list in the original partition
order. Completion order does not affect output partition ordering.

### Reuse the existing per-partition path

The original per-partition logic is moved into `_process_partition()`:

- emit `PARTITION_START`;
- wrap the Ray Dataset;
- call `_process_with_checkpointing()`;
- emit `PARTITION_COMPLETE`;
- return the partition ID and processed Ray Dataset.

The checkpoint strategy, materialization behavior, metrics, DAG monitoring,
and event types are not otherwise changed.

Calling `future.result()` also propagates a partition exception to the caller
instead of silently dropping a failed result.

### Regression coverage

Add
`tests/core/executor/test_ray_executor_partitioned_parallel.py`.

The test uses a `threading.Barrier` that can only pass when all three partition
jobs overlap. It also makes the partitions return in a concurrent execution
path and verifies that the final union order remains:

```text
[0, 1, 2]
```

The previous serial loop cannot pass the barrier and therefore fails this
test.

Additional tests cover:

- automatic planning exactly once on the driver;
- automatic and explicit Actor concurrency scaling;
- tuple/list Actor pool forms;
- global concurrency restoration after success and failure;
- explicit Actor budgets limiting partition workers;
- explicit Ray task concurrency remaining unchanged;
- automatic GPU, multi-GPU, CPU-only, and CPU-heavy partition limits;
- explicit `max_concurrent_partitions` overrides and safe fallback behavior;
- configuration parsing for the new `"auto"` default.

## Behavior before and after

```text
Before:

partition 0  [================]
partition 1                    [================]
partition 2                                      [================]
partition 3                                                        [================]

After:

partition 0  [================]
partition 1  [================]
partition 2  [================]
partition 3  [================]
```

The second timeline assumes sufficient Ray resources and no smaller explicit
Actor budget. Otherwise, the executor applies the resolved partition cap and
Ray queues tasks or Actors normally; the driver no longer adds an
unconditional serial barrier.

## Testing

The controlled benchmark harnesses and raw benchmark outputs are kept outside
the Data-Juicer repository and are not part of this PR's source diff.

### Unit test

Run:

```bash
python -m pytest -q \
  tests/core/executor/test_ray_executor_partitioned_parallel.py \
  tests/config/test_config_functions.py
```

Result:

```text
108 passed in 4.94s
```

The targeted tests cover:

- overlapping partition execution;
- result collection through completed futures;
- preservation of partition union order;
- global automatic operator planning;
- automatic and explicit Actor-pool scaling and restoration;
- resource-aware automatic partition limits and explicit overrides;
- the `max_concurrent_partitions: "auto"` configuration default.

### Controlled single-node CPU A/B benchmark

A controlled sleep Mapper was used to isolate partition-level concurrency:

- 4096 input rows;
- 4 Ray blocks;
- batch size 64;
- 100 ms of work per batch;
- one CPU worker inside each partition;
- checkpointing and automatic operator parallelism disabled;
- three repetitions per configuration.

All 18 runs completed with 4096 output rows and 4 output blocks.
The table reports median `processing_seconds`:

| Implementation | Partition | Processing time | Throughput | Speedup vs P=1 | Original / Fixed |
| --- | ---: | ---: | ---: | ---: | ---: |
| original | 1 | 11.145 s | 367.5 rows/s | 1.00x | 1.00x |
| original | 2 | 11.195 s | 365.9 rows/s | 1.00x | 1.41x |
| original | 4 | 11.398 s | 359.4 rows/s | 0.98x | 1.73x |
| fixed | 1 | 11.141 s | 367.6 rows/s | 1.00x | 1.00x |
| fixed | 2 | 7.926 s | 516.8 rows/s | 1.41x | 1.41x |
| fixed | 4 | 6.574 s | 623.1 rows/s | 1.69x | 1.73x |

P=1 changed by approximately 0.03%. At P=2 and P=4, the fix reduced
processing time by 29.20% and 42.32%, respectively.

### Initial isolated four-GPU H20 orchestration A/B benchmark

The GPU benchmark used:

- one node with 4 NVIDIA H20 GPUs;
- 100,000 input rows and 100 Ray blocks;
- batch size 1,000;
- one one-GPU Actor per partition;
- BF16 inference with a fixed token length of 64;
- local sentiment, topic-classification, and CLIP model weights;
- checkpointing, OP fusion, and automatic operator parallelism disabled;
- three repetitions per configuration.

The three model forwards were combined into one batched benchmark Mapper to
keep the 100,000-row test practical and isolate partition scheduling. Every
run produced 100,000 rows, 100 output blocks, and 100 recorded GPU batches.
All 18 runs succeeded.

Median results:

| Implementation | Partition | Processing time | Throughput | Speedup vs P=1 | Peak concurrent GPU batches |
| --- | ---: | ---: | ---: | ---: | ---: |
| original | 1 | 68.576 s | 1458.2 rows/s | 1.00x | 1 |
| original | 2 | 77.573 s | 1289.1 rows/s | 0.88x | 1 |
| original | 4 | 95.317 s | 1049.1 rows/s | 0.72x | 1 |
| fixed | 1 | 68.654 s | 1456.6 rows/s | 1.00x | 1 |
| fixed | 2 | 39.144 s | 2554.7 rows/s | 1.75x | 2 |
| fixed | 4 | 25.516 s | 3919.1 rows/s | 2.69x | 4 |

Observed GPU placement:

| Implementation | Partition | Processed rows per GPU |
| --- | ---: | --- |
| original | 2 | GPU 0: 100,000 |
| original | 4 | GPU 0: 100,000 |
| fixed | 2 | GPU 0/1: 50,000 each |
| fixed | 4 | GPU 0/1/2/3: 25,000 each |

At P=2, the fix reduced processing time by 49.54%. At P=4, it reduced
processing time by 73.23% and was 3.74x faster than the original P=4 path.

The GPU traces also show the orchestration change directly:

- original P=4 started each partition only after the previous one completed;
- fixed P=4 started all four partitions within approximately 2 ms;
- fixed P=4 partition completion times were within approximately 0.55 s;
- the four concurrent Actors were placed on four different H20 GPUs.

P=4 does not reach a full 4x speedup because the critical path still contains
approximately 10 seconds of fixed Actor creation, local model loading, CUDA
warm-up, Ray scheduling, and partition bookkeeping. The measured model-batch
work decreases from approximately 58.7 seconds at P=1 to 15.0 seconds at P=4,
while that fixed cost does not decrease.

This isolated A/B benchmark was produced while validating the core
orchestration change. Its one-Actor-per-partition setup intentionally models
per-partition placement. The final global Actor-budget and automatic
operator-planning behavior is validated separately by the end-to-end
benchmark below.

Because the final implementation interprets an explicit Actor `num_proc=1`
as a global one-Actor budget, this exact harness is retained as evidence for
the original serial barrier rather than as a performance claim for the final
explicit-Actor policy.

### Final four-GPU auto-num-proc end-to-end benchmark

The final resource-planning implementation was tested with the standard
`demos/elastic_sharding/configs/gpu_demo.yaml` operator chain:

- one node with 4 NVIDIA H20 GPUs and 8 Ray CPUs;
- Ray 2.55.1;
- 100,000 input rows and 100 Ray blocks;
- P=1, P=2, and P=4, one long run per configuration;
- separate sentiment-classification, topic-classification, and CLIP
  text-pair-similarity GPU operators;
- the original GPU batch size of 4;
- local model weights;
- every explicit operator `num_proc` removed at runtime;
- `auto_op_parallelism: true`;
- `partition.max_concurrent_partitions: "auto"`;
- checkpointing, cache, and Data-Juicer OP fusion disabled.

Every run produced and exported 100,000 rows with 100 output blocks. Runtime
instrumentation captured the global plan, the per-partition scaled plan, and
the restored plan.

Resource-plan validation:

| Partition | Global GPU num_proc | Partition workers | Per-partition GPU num_proc | Effective budget | Resolved auto limit | Trace valid |
| ---: | ---: | ---: | ---: | ---: | ---: | --- |
| 1 | 4 | 1 | 4 | 4 | 4 | yes |
| 2 | 4 | 2 | 2 | 4 | 4 | yes |
| 4 | 4 | 4 | 1 | 4 | 4 | yes |

All checks passed:

- global planning, scaling, and restoration each ran exactly once;
- all three GPU operators received the global automatic value 4;
- the automatic partition limit resolved to the four-GPU capacity;
- P=1/2/4 scaled to 4/2/1 without multiplying the global budget;
- the global values were restored after partition processing.

Performance results:

| Partition | Recipe time | Throughput | Speedup vs P=1 | Average active GPUs | Peak active GPUs |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 1698.673 s | 58.9 rows/s | 1.00x | 1.00 | 2 |
| 2 | 1258.476 s | 79.5 rows/s | 1.35x | 1.35 | 3 |
| 4 | 831.512 s | 120.3 rows/s | 2.04x | 2.07 | 4 |

P=4 reduced recipe time by 51.05% and increased throughput by 104.29%
relative to P=1. It was also 33.93% faster than P=2.

The telemetry clarifies why P=4 still helps even though every configuration
has the same logical four-GPU budget:

- P=1 primarily used GPUs 0 and 1. GPU 2 was active in only 7 of 1,568
  samples, and GPU 3 remained unused.
- P=2 used three GPUs, but GPU 3 remained unused. Its two equal 50,000-row
  partitions completed in 851.49 s and 1249.02 s, leaving a long scheduling
  tail.
- P=4 used all four GPUs evenly. Per-GPU average utilization was
  11.70%-11.75%, and its four equal partitions completed within 2.48 s of
  each other.

In the P=1 log, Ray Data commonly reported only two active GPUs out of four
requested:

```text
Active & requested resources: 2/8 CPU, 2/4 GPU
```

The execution logs also contain `ResourceBudget` backpressure. This indicates
that `num_proc=4` is an upper bound, not a guarantee that one Ray Data
execution plan will continuously occupy four GPUs. Multiple eligible stages
inside the plan share Ray Data's resource budget. Four independent
one-GPU partition pipelines used the physical cluster more consistently.

This behavior is not a recurrence of the partition resource-multiplication
bug: the captured logical plan is correct in all three cases. It is a
remaining Ray Data scheduling and pipeline-utilization characteristic. The
low absolute GPU utilization is also consistent with batch size 4, CPU
tokenization/stats/filter stages, task scheduling, and sequential model
operators.

Because this production-style benchmark is expensive and each configuration
was run once, its exact timing is less statistically robust than the
three-repeat controlled A/B benchmarks. The 51% P=4 reduction and the
per-GPU placement evidence nevertheless show a clear trend.

## Compatibility

- No configuration field is removed.
- `partition.max_concurrent_partitions` now accepts either a positive integer
  or `"auto"` and defaults to `"auto"` instead of `32`. Existing positive
  integer configurations preserve their explicit limit.
- No operator API or dataset schema is changed.
- P=1 follows the same data-processing path through a one-worker thread pool;
  CPU and GPU median benchmarks show no measurable P=1 regression.
- Processed partitions are still unioned in their original partition order.
- Existing checkpoint and event-logging functions remain the implementation
  used by every partition.
- Explicit Actor `num_proc` is now interpreted as a global job budget in the
  partitioned executor. This deliberately prevents the configured Actor pool
  from being repeated per partition.
- Explicit Ray task `num_proc` behavior is unchanged.
- Actual driver concurrency is bounded by the partition count, the resolved or
  configured partition limit, and explicit Actor budgets. Ray continues to
  enforce physical CPU, GPU, memory, and object-store resources.

## Requirements and limitations

- Automatic partition concurrency uses advertised CPU/GPU resources and
  per-worker requests. It does not predict model memory, object-store traffic,
  input skew, or Ray Data's internal per-operator resource reservation.
- A manual positive `max_concurrent_partitions` can still create many driver
  threads and Ray Data execution plans. Large overrides should be selected
  with driver, object-store, checkpoint-I/O, and model-initialization capacity
  in mind.
- Concurrent GPU partitions instantiate multiple model Actors at the same
  time up to the global Actor and partition limits. GPU memory usage therefore
  scales with the number of simultaneously admitted Actors.
- Speedup depends on partition balance, available Ray resources, operator
  cost, batch size, model-loading overhead, Ray Data backpressure, and
  object-store pressure.
- `num_proc` is a Ray Data concurrency ceiling. Internal resource reservation
  and backpressure may keep actual CPU/GPU activity below the planned value,
  as observed in the final P=1 GPU run.
- Ray 2.55.1 warns that specifying both `num_cpus` and `num_gpus` for map tasks
  is experimental. The final end-to-end recipe retains both values from the
  source recipe so each task receives a CUDA device and one CPU.
- The controlled benchmarks disable checkpointing and convergence operations
  to isolate the simple-partition execution path. The PR does not change their
  internal logic, but a large end-to-end checkpoint/resume stress test was not
  part of this validation.
- The GPU benchmark combines three model forwards in a benchmark-only Mapper;
  it validates partition scheduling and GPU placement rather than exact
  end-to-end performance of the three separate production operators.
- The final auto-num-proc benchmark does use the three separate production
  operators, but it has one repetition per partition count.
- The full repository test suite was not run.

## Validation status

- [x] Targeted executor and configuration tests, 108/108 passed
- [x] Controlled CPU A/B benchmark, 18/18 successful runs
- [x] Initial isolated 4×H20 orchestration A/B benchmark, 18/18 successful runs
- [x] Final standard-recipe 4×H20 auto-num-proc benchmark, 3/3 successful runs
- [x] Global/per-partition/restored resource-plan validation
- [x] Output row, block, and batch-count validation
- [x] GPU placement and batch-overlap validation

### DIFF
--- a/data_juicer/core/executor/ray_executor_partitioned.py
+++ b/data_juicer/core/executor/ray_executor_partitioned.py
@@ -8,6 +8,7 @@ This module implements a streamlined partitioned execution strategy for Ray mode
 5. Supports convergence points for global operations (like deduplicators)
 """
 
+import copy
 import hashlib
 import json
 import math
@@ -556,24 +557,37 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         original_concurrency = self._scale_operator_parallelism(ops, max_workers)
 
         try:
-            with ThreadPoolExecutor(max_workers=max_workers, thread_name_prefix="ray-partition") as executor:
-                futures = {
-                    executor.submit(self._process_partition, partition, i, len(partitions), ops): i
-                    for i, partition in enumerate(partitions)
-                }
-                try:
-                    for future in as_completed(futures):
-                        partition_id, processed_data = future.result()
-                        processed_partitions[partition_id] = processed_data
-                except BaseException:
-                    # Running Ray Dataset executions cannot be forcefully
-                    # interrupted here, but queued driver work should not start
-                    # after the job has already failed.
-                    for future in futures:
-                        future.cancel()
-                    raise
+            # Capture the per-partition resource plan in an isolated template.
+            # Restore the original operators before any worker starts so later
+            # convergence stages never depend on thread-pool shutdown timing.
+            isolate_operators = max_workers > 1
+            partition_ops_template = self._clone_partition_operators(ops) if isolate_operators else ops
         finally:
             self._restore_operator_parallelism(original_concurrency)
+
+        with ThreadPoolExecutor(max_workers=max_workers, thread_name_prefix="ray-partition") as executor:
+            futures = {
+                executor.submit(
+                    self._process_partition,
+                    partition,
+                    i,
+                    len(partitions),
+                    partition_ops_template,
+                    isolate_operators,
+                ): i
+                for i, partition in enumerate(partitions)
+            }
+            try:
+                for future in as_completed(futures):
+                    partition_id, processed_data = future.result()
+                    processed_partitions[partition_id] = processed_data
+            except BaseException:
+                # Running Ray Dataset executions cannot be forcefully
+                # interrupted here, but queued driver work should not start
+                # after the job has already failed.
+                for future in futures:
+                    future.cancel()
+                raise
 
         # Merge all processed partitions back into a single dataset
         logger.info("Merging processed partitions...")
@@ -588,7 +602,14 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         # Return as RayDataset wrapper
         return RayDataset(merged_dataset, cfg=self.cfg)
 
-    def _process_partition(self, partition, partition_id: int, total_partitions: int, ops: List):
+    def _process_partition(
+        self,
+        partition,
+        partition_id: int,
+        total_partitions: int,
+        ops: List,
+        isolate_operators: bool = True,
+    ):
         """Process one partition while preserving its position in the final union."""
         logger.info(f"Processing partition {partition_id + 1}/{total_partitions}")
 
@@ -599,8 +620,12 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         )
 
         try:
+            # RayDataset may temporarily mutate operator attributes while
+            # building an execution plan (for example runtime_env fallback and
+            # tracing). Give every concurrent partition its own instances.
+            partition_ops = self._clone_partition_operators(ops) if isolate_operators else ops
             partition_dataset = self._wrap_with_precomputed_parallelism(partition)
-            processed_partition = self._process_with_checkpointing(partition_dataset, partition_id, ops)
+            processed_partition = self._process_with_checkpointing(partition_dataset, partition_id, partition_ops)
         except Exception as error:
             self.log_partition_failed(partition_id, str(error), retry_count=0)
             raise
@@ -612,6 +637,17 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         )
 
         return partition_id, processed_partition.data
+
+    @staticmethod
+    def _clone_partition_operators(ops: List) -> List:
+        """Create an isolated operator graph for one partition."""
+        try:
+            return copy.deepcopy(ops)
+        except Exception as error:
+            raise RuntimeError(
+                "Failed to isolate operators for concurrent partition execution. "
+                "All partition operators must support deep copying."
+            ) from error
 
     def _configure_operator_parallelism(self, ops: List) -> None:
         """Plan global actor resources once before partition threads.
--- a/data_juicer/core/executor/ray_executor_partitioned.py
+++ b/data_juicer/core/executor/ray_executor_partitioned.py
@@ -10,6 +10,7 @@ This module implements a streamlined partitioned execution strategy for Ray mode
 
 import hashlib
 import json
+import math
 import os
 import shutil
 import time
@@ -456,6 +457,14 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         if self.partition_mode == "auto":
             self._configure_auto_partitioning(dataset, ops)
 
+        # Calculate automatic Ray operator parallelism once on the driver.
+        # Every partition shares the same operator objects, so recalculating in
+        # concurrent partition threads would mutate their resource settings
+        # concurrently. Partition execution scales the resulting global actor
+        # pool sizes below, while Ray task pools can continue to use their
+        # resource-driven default when num_proc is None.
+        self._configure_auto_operator_parallelism(ops)
+
         # Initialize DAG execution planning with final partition count
         # Pass ops to avoid redundant loading
         self._initialize_dag_execution(self.cfg, ops=ops)
@@ -542,15 +551,27 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
             "driver threads and checkpointing support..."
         )
         processed_partitions = [None] * len(partitions)
+        original_auto_concurrency = self._scale_auto_operator_parallelism(ops, max_workers)
 
-        with ThreadPoolExecutor(max_workers=max_workers, thread_name_prefix="ray-partition") as executor:
-            futures = {
-                executor.submit(self._process_partition, partition, i, len(partitions), ops): i
-                for i, partition in enumerate(partitions)
-            }
-            for future in as_completed(futures):
-                partition_id, processed_data = future.result()
-                processed_partitions[partition_id] = processed_data
+        try:
+            with ThreadPoolExecutor(max_workers=max_workers, thread_name_prefix="ray-partition") as executor:
+                futures = {
+                    executor.submit(self._process_partition, partition, i, len(partitions), ops): i
+                    for i, partition in enumerate(partitions)
+                }
+                try:
+                    for future in as_completed(futures):
+                        partition_id, processed_data = future.result()
+                        processed_partitions[partition_id] = processed_data
+                except BaseException:
+                    # Running Ray Dataset executions cannot be forcefully
+                    # interrupted here, but queued driver work should not start
+                    # after the job has already failed.
+                    for future in futures:
+                        future.cancel()
+                    raise
+        finally:
+            self._restore_auto_operator_parallelism(original_auto_concurrency)
 
         # Merge all processed partitions back into a single dataset
         logger.info("Merging processed partitions...")
@@ -576,7 +597,7 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         )
 
         try:
-            partition_dataset = RayDataset(partition, cfg=self.cfg)
+            partition_dataset = self._wrap_with_precomputed_parallelism(partition)
             processed_partition = self._process_with_checkpointing(partition_dataset, partition_id, ops)
         except Exception as error:
             self.log_partition_failed(partition_id, str(error), retry_count=0)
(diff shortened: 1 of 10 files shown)


fengrui-z:
Additional changes pushed in `978012845e`:

- Added bounded partition concurrency and failure cleanup.
- Automatic Ray parallelism is now calculated once and shared across concurrent partitions; explicit `num_proc` settings remain unchanged.
- Expanded coverage to 12 tests for concurrency, events, checkpoints, and resource restoration.

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -531,35 +543,35 @@ def _process_with_simple_partitioning(self, dataset: RayDataset, ops: List):
...
-            # Log partition completion event
-            self._log_event(
-                event_type=EventType.PARTITION_COMPLETE,
-                message=f"Completed processing of partition {i+1}/{len(partitions)}",
-                partition_id=i,
-            )
+        try:
+            with ThreadPoolExecutor(max_workers=max_workers, thread_name_prefix="ray-partition") as executor:
+                futures = {
+                    executor.submit(self._process_partition, partition, i, len(partitions), ops): i
```
cmgzn: Non-blocking issue

cmgzn (commented):
Concurrency review — three shared-mutable-state issues in the new ThreadPoolExecutor path.

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -574,6 +586,95 @@ def _process_with_simple_partitioning(self, dataset: RayDataset, ops: List):
...
+
+        self._log_event(
+            event_type=EventType.PARTITION_START,
+            message=f"Starting processing of partition {partition_id + 1}/{total_partitions}",
+            partition_id=partition_id,
+        )
+
+        try:
+            partition_dataset = self._wrap_with_precomputed_parallelism(partition)
+            processed_partition = self._process_with_checkpointing(partition_dataset, partition_id, ops)
```
cmgzn: `_process_with_checkpointing` calls `_pre_execute_operations_with_dag_monitoring`, which calls `_mark_dag_node_started` — that writes `self.current_dag_node = node_id` (dag_execution_mixin.py:255) without synchronization. This scalar is shared across all partition threads.

**Reproduced:** partitions 0/1/2 each write their own node_id, but read back `"op_0_partition_3"`. `pipeline_dag.mark_node_started` also overwrites `node["start_time"]`, corrupting `actual_duration` when multiple threads hit the same node.

**Suggestion:** DAG monitoring in the concurrent path needs a lock around `current_dag_node` and node state mutations, or `current_dag_node` should become a per-partition dict.

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -531,35 +543,35 @@ def _process_with_simple_partitioning(self, dataset: RayDataset, ops: List):
...
+                        processed_partitions[partition_id] = processed_data
+                except BaseException:
+                    # Running Ray Dataset executions cannot be forcefully
+                    # interrupted here, but queued driver work should not start
+                    # after the job has already failed.
+                    for future in futures:
+                        future.cancel()
+                    raise
+        finally:
+            self._restore_auto_operator_parallelism(original_auto_concurrency)
```
cmgzn: Non-blocking issue

Qirui-jiao:
`ed10a77` updates:

-  Explicit actor num_proc values are now treated as global resource budgets and distributed across concurrent partitions, preventing each partition from requesting a complete actor pool independently.
- The default max_concurrent_partitions has been changed from a fixed value of 32 to resource-aware auto, which determines a safe concurrency limit based on Ray cluster resources and operator CPU/GPU requirements.

fengrui-z (approved):
LGTM

data_juicer/core/executor/ray_executor_partitioned.py

data_juicer/core/executor/ray_executor_partitioned.py

data_juicer/core/executor/ray_executor_partitioned.py

PR #1024, opened 2026-07-30
Use public Ray compute strategies

fengrui-z:
## Summary

Replace Ray’s private `get_compute_strategy` helper with the public `TaskPoolStrategy` and `ActorPoolStrategy` APIs.

## Changes

- Use `ActorPoolStrategy(size=op.num_proc)` for actor-based Mapper and Filter operations.
- Use `TaskPoolStrategy(size=op.num_proc)` for task-based operations.
- Add regression coverage for Mapper/Filter × task/actor combinations.

## Motivation

The previous implementation imported `get_compute_strategy` from `ray.data._internal`, which is not a stable public API. It also passed the deprecated `concurrency` argument internally.

The new implementation follows Ray’s public API while preserving the existing fixed worker-count behavior.

## Validation

- 20 related tests passed.
- Verified end-to-end execution with Ray 2.52.0.
- Confirmed task mode uses `TaskPoolMapOperator`.
- Confirmed actor mode uses `ActorPoolMapOperator`.
- All pre-commit checks passed.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -8,7 +8,7 @@ import pyarrow
 import ray
 from jsonargparse import Namespace
 from loguru import logger
-from ray.data._internal.util import get_compute_strategy
+from ray.data import ActorPoolStrategy, TaskPoolStrategy
 
 from data_juicer.core.data import DJDataset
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM

PR #1025, opened 2026-07-30
perf: Reduce eager Ray dataset actions

fengrui-z:
## Summary

Reduce unnecessary eager Ray Dataset actions and fix row retrieval truncation.

## Changes

- Remove the full `count()` call before building an operator pipeline.
- Use `take(k)` directly instead of `count()`, `limit(k)`, and `take()`.
- Use `take_all()` when `get_column()` is expected to return every row.
- Avoid repeated `columns()` and `schema()` calls.
- Add regression coverage with datasets containing more than 20 rows.

## Motivation

Ray Datasets are lazily evaluated, but `count()` triggers execution. Calling it before constructing every processing pipeline adds unnecessary work, especially when the partitioned executor invokes `process()` for each partition or checkpoint group.

Additionally, Ray’s `take()` defaults to 20 rows. The previous implementation could therefore silently truncate results from `get()` and `get_column()` even when more rows were requested.

## Validation

- 20 related tests passed.
- Verified that 25 requested rows are returned without truncation.
- Verified that `process()` no longer calls `count()` before building the execution plan.
- All pre-commit checks passed.

## Note

`get_column(k=None)` now follows its documented behavior and returns all rows using `take_all()`. As before, callers should avoid using it with datasets too large to fit in driver memory.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -112,10 +112,14 @@ class RayDataset(DJDataset):
         Returns:
             Schema: Dataset schema containing column names and types
         """
-        if self.data is None or self.data.columns() is None:
+        if self.data is None:
             raise ValueError("Dataset is empty or not initialized")
 
-        return Schema.from_ray_schema(self.data.schema())
+        ray_schema = self.data.schema()
+        if ray_schema is None:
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM

PR #1032, opened 2026-08-05
fix(ray): preserve elastic actor pool concurrency

fengrui-z:
## Summary

- Add a shared `_build_actor_pool_strategy()` helper for Ray Actor operators.
- Preserve elastic concurrency ranges by converting `(min_proc, max_proc)` to:
  `ActorPoolStrategy(min_size=min_proc, max_size=max_proc)`.
- Keep fixed integer concurrency unchanged with `ActorPoolStrategy(size=num_proc)`.
- Apply the conversion in both RayDataset Actor paths:
  - Mapper execution
  - Filter statistics execution
- Add regression coverage for elastic ActorPool concurrency.

## Root Cause

`calculate_ray_np()` may assign Ray Actor operators an elastic concurrency range such as `(1, 30)`. Previously, `RayDataset` passed this tuple directly to:

```python
ActorPoolStrategy(size=op.num_proc)
```

Ray expects `size` to be an integer, so this caused:

```text
TypeError: '<' not supported between instances of 'tuple' and 'int'
```

before any Actor, model, or GPU workload was started.

## Fix

The new helper preserves the planner’s intended semantics:

```python
# Fixed concurrency
ActorPoolStrategy(size=4)

# Elastic concurrency
ActorPoolStrategy(min_size=1, max_size=30)
```

This prevents the tuple type error without collapsing elastic ActorPool concurrency into a fixed maximum-sized pool.

## Validation

Passed:

- `tests/core/data/test_ray_dataset.py`: 17 passed
- `tests/core/executor/test_ray_executor.py`: 6 passed
- `tests/core/executor/test_ray_executor_partitioned.py`: 24 passed
- `tests/core/executor/test_ray_executor_partitioned_parallel.py`: 29 passed
- `tests/core/executor/test_partitioned_integration.py`: 13 passed, 1 skipped

`test_ray_exporter.py` has a pre-existing failure in `test_lance_keep_hashes`: it also reproduces on the unmodified base commit (`753debf`) and is caused by the current Lance 7.0.0 / Ray Data exporter compatibility issue.

### DIFF
--- a/data_juicer/core/data/ray_dataset.py
+++ b/data_juicer/core/data/ray_dataset.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from functools import partial
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
 import pyarrow
 import ray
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM

Qirui-jiao (approved):

PR #1036, opened 2026-08-05
[Bugfix]: Preserve partition checkpoints across Ray block-layout changes

### LINKED ISSUE #1030: 分区续跑时，续跑时的分区切分结果会与第一次不同，导致检查点无法生效
### Before Asking

- [x] I have asked the [Data-Juicer Q&A Copilot](https://datajuicer.github.io/data-juicer/en/main/docs_index.html) (available on Doc Site, DingTalk, or Discord), but the problem still persists.

- [x] I have read the [README](https://github.com/datajuicer/data-juicer/blob/main/README.md) carefully.

- [x] I have pulled the latest code of main branch to run again and the problem still existed.

### Search before asking

- [x] I have searched the Data-Juicer [issues](https://github.com/datajuicer/data-juicer/issues) and found no similar questions.

### Question

如题，我用分区模式执行过滤任务，共切分为32个分区，第一次执行在第12个分区中断，第二次续跑时，会重新分区，然后检查新的分区与检查点不一致，会把检查点全部删除然后重跑，期间没有动过配置，完全相同的配置续跑时会出现这个问题
日志如下
data_juicer.core.executor.ray_executor_partitioned:964 - Created 32 partitions
2026-08-05 10:44:48.045 | INFO     | data_juicer.core.executor.ray_executor_partitioned:968 - Found existing partitioning info, validating...
2026-08-05 10:44:48.048 | ERROR    | data_juicer.core.executor.ray_executor_partitioned:926 - Partition 0 row count mismatch: current=899772, saved=1008747
2026-08-05 10:44:48.048 | WARNING  | data_juicer.core.executor.ray_executor_partitioned:973 - Partition validation FAILED - partitions don't match saved info. This can happen if the input data changed or Ray's internal state differs. Clearing checkpoints and starting fresh.
2026-08-05 10:44:48.049 | WARNING  | data_juicer.core.executor.ray_executor_partitioned:1006 - Clearing invalid checkpoints in /share/airesearch/slm/slm_dataset_scored/OpenBMB__Ultra-FineWeb/ppl_qwen25_2_baseop2mindedup2llmConsistency/20260803_065731_99d69b/checkpoints
2026-08-05 10:44:50.767 | INFO     | data_juicer.core.executor.ray_executor_partitioned:982 - Collecting partition metadata for checkpoint validation...

### Environment Information

_No response_

### Additional Context

_No response_

Qirui-jiao:
## Summary

This PR makes `ray_partitioned` jobs explicitly resumable even when Ray produces a different physical block layout between two otherwise identical runs.

It introduces:

- a new `--resume <job_id>` CLI option for `ray_partitioned` jobs;
- persisted inclusive/exclusive row offsets for every logical partition;
- a stable full-content hash for every partition;
- reconstruction of the original logical partitions with `split_at_indices()`;
- fail-safe validation before any existing checkpoint is reused;
- a printed resume token for new checkpoint-enabled partitioned jobs.

Example:

```bash
# Initial run. The log prints: "Resume token: <job_id>"
dj-process --config recipe.yaml

# Resume the same job without changing the recipe
dj-process --config recipe.yaml --resume <job_id>
```

## Motivation

`Dataset.split(n)` depends on the physical Ray block layout. A job can therefore produce different logical partitions after a restart even when both the input data and recipe are unchanged.

For example, the regression tests force the same input to be read as 64 blocks on the first run and 65 blocks on the second run:

```text
first run:  partition 0 = 256 rows
second run: partition 0 = 380 rows
```

The existing resume path detects this row-count mismatch, deletes all partition checkpoints, and restarts the whole job. This makes checkpointing ineffective for long-running partitioned pipelines whose Ray block layout changes across processes or cluster states.

### Persist stable partition metadata

After the initial split, `partitioning_info.json` now stores the following for each partition:

```json
{
  "partition_id": 0,
  "row_count": 256,
  "start_row": 0,
  "end_row": 256,
  "content_hash": "..."
}
```

The file also records the hash algorithm identifier:

```text
sha256-multiset-v1
```

The content hash is built from canonical per-row SHA-256 digests and combined with order- and Ray-batch-boundary-independent accumulators. This allows the same logical partition to validate even when Ray changes its internal physical blocks.

### Recreate the original logical partitions

When `--resume <job_id>` is provided, the executor:

1. locates the original work and checkpoint directories;
2. validates that the current configuration matches the original run;
3. loads `partitioning_info.json`;
4. reuses the saved partition count, including when partition mode is `auto`;
5. enables ordered Ray execution before loading the dataset;
6. recreates the original partitions from the saved row offsets with `split_at_indices()`;
7. recomputes and validates every partition content hash;
8. reuses existing checkpoints only after all validation succeeds.

### Fail closed without deleting checkpoints

Explicit resume fails with an actionable error and leaves existing checkpoints untouched when:

- the job or work directory cannot be located;
- `partitioning_info.json` is missing;
- the saved partition metadata is incomplete;
- the hash algorithm is unsupported;
- row boundaries are invalid;
- the current input content does not match the saved hashes.

This differs intentionally from the legacy `--job_id` behavior, which is retained for backward compatibility and may clear mismatched checkpoints before starting fresh.

## CLI behavior and compatibility

- `--resume` is supported only by the `ray_partitioned` executor.
- `--resume ID --job_id ID` is accepted when both IDs are identical.
- Conflicting `--resume` and `--job_id` values are rejected.
- The resume-only CLI argument is excluded from configuration equality checks, so the original recipe remains unchanged.
- Existing metadata can still be deserialized with empty default values for the new fields.
- Metadata created by older versions does not contain the full content hashes and row offsets required for safe explicit resume. In that case, `--resume` refuses to continue and preserves the existing checkpoint files.
- Existing non-explicit `--job_id` behavior remains available.

## Scope

The change is limited to partition checkpoint resumption:

- no recipe changes are required;
- the normal initial partitioning algorithm remains unchanged;
- legacy `--job_id` behavior remains unchanged;
- complete partition hashes are collected only when checkpointing is enabled;
- no unrelated operator or export behavior is modified.

### Unit tests

```bash
NUMBA_DISABLE_JIT=1 PYTHONDONTWRITEBYTECODE=1 \
python -m pytest -q -p no:cacheprovider \
  tests/core/executor/test_ray_partition_resume.py \
  tests/config/test_config_functions.py::ResolveJobIdTest \
  tests/config/test_config_functions.py::BuildBaseParserTest::test_has_required_args
```

Result:

```text
17 passed in 3.63s
```

The tests cover:

- `--resume` ID resolution and CLI validation;
- conflicting `--resume`/`--job_id` values;
- rejection on non-partitioned executors;
- content-hash stability across Arrow batch boundaries and batch order;
- detection of changed partition content;
- persistence of row offsets and content hashes;
- reconstruction with saved row counts and explicit offsets;
- invalid boundary rejection;
- preservation of checkpoints after explicit-resume hash mismatch;
- safe rejection of legacy metadata without hashes or boundaries.

### CPU end-to-end regression

Configuration:

- 8,192 deterministic rows;
- 32 partitions;
- `text_length_filter`;
- first run: 64 Ray blocks, interrupted at partition 12;
- resume run: 65 Ray blocks with the same input and recipe.

Legacy behavior reproduced:

```text
Partition 0 row count mismatch: current=380, saved=256
Clearing invalid checkpoints
```

Fixed behavior with `--resume`:

- recreated all 32 partitions at the original 256-row boundaries;
- validated all 32 content hashes;
- reused all 11 completed checkpoints without changing their mtimes;
- preserved the original `partitioning_info.json` SHA-256;
- produced exactly 8,192 unique IDs with no missing or duplicated rows.

### Four-GPU end-to-end regression

Configuration:

- 4×NVIDIA H20;
- 512 deterministic rows;
- 32 partitions with 4 concurrent partition pipelines;
- `llm_perplexity_filter` with `Qwen/Qwen2.5-0.5B`;
- one Ray actor and one GPU per concurrent partition;
- first run: 64 Ray blocks, interrupted at partition 12;
- resume run: 65 Ray blocks with the same input and recipe.

All four Ray GPU actors received distinct GPU IDs (`0`, `1`, `2`, and `3`) and imported Data-Juicer from this PR checkout.

Legacy behavior reproduced:

```text
Partition 0 row count mismatch: current=24, saved=16
Clearing invalid checkpoints
```

Fixed behavior with `--resume`:

- recreated the original 32×16-row partitions;
- validated all saved content hashes;
- reused all 11 completed GPU checkpoints without changing their mtimes;
- did not log `Clearing invalid checkpoints`;
- produced 32 final checkpoints;
- produced exactly 512 unique IDs in the range `0..511`;
- produced finite `llm_perplexity` values for all 512 rows;
- completed without CUDA OOM or unexpected exceptions.

## Performance note

Computing complete content hashes adds one validation scan when partition metadata is first created and another scan during explicit resume. This is intentional: checkpoint reuse is allowed only after the complete saved partition contents have been verified.

The synthetic tests also showed that reading many small Parquet checkpoints can increase the number of Ray blocks and final output shards. This affects merge/export overhead for very small test partitions, but it is independent of the checkpoint-correctness fix and is intentionally left for a separate optimization.

## Checklist

- [x] Added explicit `--resume <job_id>` CLI support.
- [x] Preserved existing `--job_id` behavior.
- [x] Persisted partition row boundaries and complete content hashes.
- [x] Added fail-safe validation that does not delete checkpoints.
- [x] Added unit coverage for the new behavior and compatibility checks.
- [x] Documented explicit resume behavior and usage in the English and Chinese partition/checkpoint guides.
- [x] Reproduced the original issue on CPU and four GPUs.
- [x] Verified checkpoint reuse and exact final output on CPU and GPU.

### DIFF
--- a/data_juicer/core/executor/ray_executor_partitioned.py
+++ b/data_juicer/core/executor/ray_executor_partitioned.py
@@ -40,8 +40,9 @@ from data_juicer.utils.lazy_loader import LazyLoader
 ray = LazyLoader("ray")
 
 _AUTO_CPU_PARTITION_CAP = 4
-_PARTITION_CONTENT_HASH_ALGORITHM = "sha256-multiset-v1"
+_PARTITION_CONTENT_HASH_ALGORITHM = "sha256-sequence-v1"
 _PARTITION_CONTENT_HASH_MODULUS = 1 << 256
+_PARTITION_CONTENT_HASH_BASE = int.from_bytes(hashlib.sha256(b"data-juicer-partition-sequence-v1").digest(), "big") | 1
 
 
 def _canonical_row_bytes(row: Dict) -> bytes:
@@ -54,37 +55,36 @@ def _canonical_row_bytes(row: Dict) -> bytes:
 
 
 def _hash_partition_batch(batch):
-    """Return an order-independent hash accumulator for one Arrow batch."""
+    """Return an order-sensitive rolling hash for one Arrow batch."""
     import pyarrow
 
-    digest_sum = 0
-    digest_xor = 0
+    sequence_hash = 0
     rows = batch.to_pylist()
     for row in rows:
         digest = int.from_bytes(hashlib.sha256(_canonical_row_bytes(row)).digest(), "big")
-        digest_sum = (digest_sum + digest) % _PARTITION_CONTENT_HASH_MODULUS
-        digest_xor ^= digest
+        sequence_hash = (sequence_hash * _PARTITION_CONTENT_HASH_BASE + digest) % _PARTITION_CONTENT_HASH_MODULUS
 
     return pyarrow.table(
         {
             "row_count": [len(rows)],
-            "digest_sum": [f"{digest_sum:064x}"],
-            "digest_xor": [f"{digest_xor:064x}"],
+            "sequence_hash": [f"{sequence_hash:064x}"],
         }
     )
 
 
 def _combine_partition_hash_partials(partials: List[Dict]) -> tuple:
-    """Combine batch accumulators into a stable partition content hash."""
+    """Combine ordered batch hashes without depending on batch boundaries."""
     row_count = 0
-    digest_sum = 0
-    digest_xor = 0
+    sequence_hash = 0
     for partial in partials:
-        row_count += int(partial["row_count"])
-        digest_sum = (digest_sum + int(partial["digest_sum"], 16)) % _PARTITION_CONTENT_HASH_MODULUS
-        digest_xor ^= int(partial["digest_xor"], 16)
+        partial_count = int(partial["row_count"])
+        sequence_hash = (
+            sequence_hash * pow(_PARTITION_CONTENT_HASH_BASE, partial_count, _PARTITION_CONTENT_HASH_MODULUS)
+            + int(partial["sequence_hash"], 16)
+        ) % _PARTITION_CONTENT_HASH_MODULUS
+        row_count += partial_count
 
-    payload = f"{_PARTITION_CONTENT_HASH_ALGORITHM}:" f"{row_count}:{digest_sum:064x}:{digest_xor:064x}"
+    payload = f"{_PARTITION_CONTENT_HASH_ALGORITHM}:{row_count}:{sequence_hash:064x}"
     return hashlib.sha256(payload.encode("ascii")).hexdigest(), row_count
 
 
@@ -1277,9 +1277,10 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
     ) -> PartitionMetadata:
         """Collect metadata from a partition for validation on resume.
 
-        The complete content hash is independent of Ray block boundaries and
-        row order within the partition. It is used for strict explicit resume
-        validation; the first-row hash remains for backward compatibility.
+        The complete content hash is independent of Ray block boundaries but
+        sensitive to row order within the partition. It is used for strict
+        explicit resume validation; the first-row hash remains for backward
+        compatibility.
         """
         if compute_content_hash:
             content_hash, row_count = self._compute_partition_content_hash(partition)
@@ -1308,7 +1309,9 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
         )
 
     def _compute_partition_content_hash(self, partition) -> tuple:
-        """Compute a block-boundary-independent hash of a partition."""
+        """Compute an order-sensitive, block-boundary-independent hash."""
+        # preserve_order is enabled before the source Dataset is created, so
+        # take_all() returns these one-row partials in partition row order.
         partials = partition.map_batches(_hash_partition_batch, batch_format="pyarrow").take_all()
         return _combine_partition_hash_partials(partials)
 
@@ -1397,9 +1400,6 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
                 f"expected={saved_info.num_partitions}, actual={len(saved_info.partitions)}"
             )
 
-        if saved_info.num_partitions == 1:
-            return [dataset.data.materialize()]
-
         split_indices = []
         expected_start = 0
         for partition_id, meta in enumerate(saved_info.partitions):
@@ -1420,6 +1420,9 @@ class PartitionedRayExecutor(ExecutorBase, DAGExecutionMixin, EventLoggingMixin)
                 "Saved row boundaries do not cover the saved total row count: "
                 f"boundaries_end={expected_start}, total_rows={saved_info.total_rows}"
             )
+
+        if saved_info.num_partitions == 1:
+            return [dataset.data.materialize()]
 
         logger.info(f"Recreating partitions at saved row boundaries: {split_indices}")
         return dataset.data.split_at_indices(split_indices)
--- a/data_juicer/core/executor/ray_executor_partitioned.py
+++ b/data_juicer/core/executor/ray_executor_partitioned.py
@@ -40,6 +40,52 @@ from data_juicer.utils.lazy_loader import LazyLoader
 ray = LazyLoader("ray")
 
 _AUTO_CPU_PARTITION_CAP = 4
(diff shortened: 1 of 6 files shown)


data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -40,6 +40,52 @@
...
+_PARTITION_CONTENT_HASH_MODULUS = 1 << 256
+
+
+def _canonical_row_bytes(row: Dict) -> bytes:
+    """Serialize one row consistently for partition content hashing."""
+    try:
+        serialized = json.dumps(row, sort_keys=True, default=str, separators=(",", ":"), ensure_ascii=False)
+    except Exception:
+        serialized = str(row)
+    return serialized.encode("utf-8")
```
fengrui-z, line 54: Partition hashing ignores row order and may reuse stale checkpoints: 
uses sum/XOR accumulators, so [A, B, C] and [A, C, B] produce the same hash. If the first row remains unchanged, the existing first-row check also misses it. This can incorrectly accept reordered input and reuse checkpoints. Include row positions in the hash, or explicitly guarantee order-insensitive pipelines.

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -1275,6 +1384,46 @@ def _validate_partitions(self, partitions: List, saved_info: PartitioningInfo) -
...
+                f"current={self.num_partitions}, saved={saved_info.num_partitions}"
+            )
+        if len(saved_info.partitions) != saved_info.num_partitions:
+            raise RuntimeError(
+                "Saved partition metadata is incomplete: "
+                f"expected={saved_info.num_partitions}, actual={len(saved_info.partitions)}"
+            )
+
+        if saved_info.num_partitions == 1:
+            return [dataset.data.materialize()]
```
fengrui-z: Boundary validation is skipped for single-partition jobs:

ray_executor_partitioned.py#L1398-L1399 returns early when num_partitions == 1, without validating start_row, end_row, or total_rows. Corrupted metadata could therefore still pass hash/row-count checks and reuse checkpoints, contrary to the fail-closed design. Move boundary validation before the early return.

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -40,6 +40,52 @@
...
+
+
+def _hash_partition_batch(batch):
+    """Return an order-sensitive rolling hash for one Arrow batch."""
+    import pyarrow
+
+    sequence_hash = 0
+    rows = batch.to_pylist()
+    for row in rows:
+        digest = int.from_bytes(hashlib.sha256(_canonical_row_bytes(row)).digest(), "big")
```
cmgzn, line 64: Per-row json.dumps + sha256 in _hash_partition_batch could be costly at scale, is there a faster way to compute this?

data_juicer/core/executor/ray_executor_partitioned.py
cmgzn, line 1437: Nit: `_enable_deterministic_execution()` is already called in `_run_impl()` (L516) before `load_dataset()`, so this second call is redundant in the current call graph. Fine to keep as defensive coding if `_split_dataset_deterministic` is meant to be callable standalone.

cmgzn (approved):
LGTM — no blocking issues.

data_juicer/core/executor/ray_executor_partitioned.py
Qirui-jiao: Fixed.

data_juicer/core/executor/ray_executor_partitioned.py

data_juicer/core/executor/ray_executor_partitioned.py
Qirui-jiao: Benchmarking shows that most of the overhead comes from converting Arrow data to Python objects and serializing each row to JSON. Replacing SHA-256 with xxHash improved performance by only about 6%, while introducing a new hash protocol and weaker collision resistance. A more effective Arrow-native implementation would require substantially broader changes and should be handled in a separate performance-focused PR. The current implementation prioritizes order sensitivity, stability across Ray batch boundaries, and fail-
  closed validation, so this PR should remain focused on the checkpoint resume correctness issue.

PR #1038, opened 2026-08-06
feat(selector): add image OHEM selector

nofear-spec:
## Motivation

Add a generic image OHEM selector for retaining high-loss image samples.

## Changes

- Add `image_ohem_selector`
- Support user-defined model factory and per-sample loss function
- Support `top_ratio` and `topk`
- Add documentation and unit tests

## Algorithm

For each image sample, the user-defined scoring function computes an individual training loss. After scoring the dataset, samples are sorted by loss in descending order. The selector keeps the specified top-k or top-ratio samples and filters out low-loss samples. The computed losses are also cached in the sample statistics for later analysis or reuse.

## Testing

- `python3 -m py_compile ...`
- `git diff --check`

### DIFF
--- a/data_juicer/config/config_all.yaml
+++ b/data_juicer/config/config_all.yaml
@@ -859,6 +859,15 @@ process:
       words_aug_join_char: ""                                 # the join char between words to augment
   - general_field_filter:                                    # Filter to keep samples based on a general field filter condition.
       filter_condition: ""                                  # The filter condition as a string. It can include logical operators (and/or) and chain comparisons.
+  - image_ohem_selector:                                   # select high-loss image samples using a user-defined model/loss.
+      score_file: ""                                        # Python file defining score_fn(model, samples, images, device, **kwargs)
+      score_function: score_fn
+      model_function: model_factory
+      top_ratio: 0.3                                        # retain this fraction of highest-loss samples
+      topk: null                                             # optional maximum number of selected samples
+      batch_size: 8
+      loss_field: image_ohem_loss
+      device: auto
   - image_aesthetics_filter:                                # filter samples according to the aesthetics score of images.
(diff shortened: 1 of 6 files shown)


Qirui-jiao:
Thank you for contributing the image_ohem_selector. The overall idea is useful. Before merging, could you please address the following issues?

  1. Model lifecycle / GPU memory

  > The model created by model_factory is stored in self._model, but free_models() only releases models registered in the global MODEL_ZOO. Therefore, the user-created model remains referenced after this selector finishes and
  > may keep its GPU memory allocated, potentially causing OOMs in subsequent operators. Please release it in a finally block after scoring, for example by moving it to CPU when possible and setting self._model = None.

  2. Embedded image bytes

  > Please align image loading with Data-Juicer’s existing multimodal conventions. image_bytes_key currently defaults to "", which disables the standard image_bytes field. Also, _images() treats the byte list as all-or-nothing
  > and fails when individual entries are None. Please default to "image_bytes" and use load_mm_bytes_from_sample() per image, falling back to the image path when no valid bytes are available. Tests should cover both bytes-only
  > input and partially missing byte entries.

  3. Dynamic module loading

  > The module created from score_file should be registered in sys.modules before calling exec_module(). Without this, valid Python files using patterns such as from __future__ import annotations together with @dataclass can
  > fail during import. Please use a unique module name, register it before execution, and remove it on failure. load_custom_operators() in config.py provides an existing implementation pattern.

nofear-spec:
Thanks for the careful review and suggestions. I addressed the issues in the latest commit:

Release the user model in a finally block, including error paths.
Use image_bytes by default and fall back to image paths per image.
Register dynamically loaded modules in sys.modules with unique names and clean them up on failure.
Add tests for model cleanup, image bytes fallback, and dynamic module loading.
Expand the usage documentation.
All 11 related unit tests and pre-commit checks pass locally.

As follow-up work, I plan to explore a local/Ray-compatible loss mapper and a distributed top-k implementation for large-scale datasets.

Qirui-jiao (approved):

PR #1051, opened 2026-08-25
fix(deduplicator): handle empty token set in document_minhash_deduplicator

cmgzn:
## Problem

`document_minhash_deduplicator.compute_hash()` raises `ValueError: zero-size array to reduction operation minimum which has no identity` when a sample's text produces zero tokens (empty string, or text shorter than `window_size`).

## Root Cause

In `compute_hash()` (line 264), when `len(hv) == 0`, the code attempts:
```python
phv = np.bitwise_and((hv[:, None] * self.perm_a + self.perm_b) % MERSENNE_PRIME, MAX_HASH)
hash_values = phv.min(axis=0)  # ValueError on empty array
```

## Fix

When the token hash vector is empty, fill `hash_values` with `MAX_HASH` (indicating no signature) instead of attempting a min-reduction on an empty array. This makes the sample effectively unique in deduplication (max-hash signature won't collide with any real content).

## Reproduction

```python
import numpy as np
hv = np.array([], dtype=np.uint64)
phv = np.bitwise_and((hv[:, None] * perm_a + perm_b) % MERSENNE_PRIME, MAX_HASH)
phv.min(axis=0)  # → ValueError
```

## Tests

Added `DocumentMinhashDeduplicatorEmptyInputTest` with two cases:
- Empty text string
- Text shorter than window_size (produces zero tokens)

---
Tracked in: cmgzn/data-juicer#153

### DIFF
--- a/data_juicer/ops/deduplicator/document_minhash_deduplicator.py
+++ b/data_juicer/ops/deduplicator/document_minhash_deduplicator.py
@@ -262,8 +262,7 @@ class DocumentMinhashDeduplicator(Deduplicator):
         # compute minhash value
         hv = np.fromiter((sha1_hash32(token) for token in tokens), dtype=np.uint64, count=len(tokens))
         if len(hv) == 0:
-            phv = np.bitwise_and((hv[:, None] * self.perm_a + self.perm_b) % MERSENNE_PRIME, MAX_HASH)
-            hash_values = phv.min(axis=0)
+            hash_values = np.full(self.num_permutation, MAX_HASH, dtype=np.uint64)
         else:
             bytes_per_token = self.num_permutation * np.dtype(np.uint64).itemsize
             tokens_per_block = max(1, _MINHASH_PERMUTATION_BLOCK_BYTES // bytes_per_token)
--- a/tests/ops/deduplicator/test_document_minhash_deduplicator.py
+++ b/tests/ops/deduplicator/test_document_minhash_deduplicator.py
@@ -1053,15 +1053,12 @@ class DocumentMinhashDeduplicatorRepeatedCallTest(DataJuicerTestCaseBase):
                          "UID-based dedup must not carry state between process() calls")
 
 
-if __name__ == '__main__':
-    unittest.main()
-
-
 class DocumentMinhashDeduplicatorEmptyInputTest(DataJuicerTestCaseBase):
-    """Regression test for empty-token ValueError (GitHub issue).
+    """Tests for samples whose text produces zero tokens.
 
-    When a sample's text produces zero tokens (e.g. empty string or text
-    shorter than window_size), compute_hash should not raise ValueError.
+    An empty string, or text shorter than window_size, yields no shingles.
+    compute_hash must still produce a well-formed minhash signature for such
+    samples instead of failing on the empty token array.
     """
 
     def test_empty_text_no_crash(self):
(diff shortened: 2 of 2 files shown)


data_juicer/ops/deduplicator/document_minhash_deduplicator.py
```diff
@@ -262,8 +262,7 @@ def compute_hash(self, sample):
         # compute minhash value
         hv = np.fromiter((sha1_hash32(token) for token in tokens), dtype=np.uint64, count=len(tokens))
         if len(hv) == 0:
-            phv = np.bitwise_and((hv[:, None] * self.perm_a + self.perm_b) % MERSENNE_PRIME, MAX_HASH)
-            hash_values = phv.min(axis=0)
+            hash_values = np.full(self.num_permutation, MAX_HASH, dtype=np.uint64)
```
fengrui-z, line 265: The PR description says the max-hash signature makes the sample "effectively unique", but two empty samples produce identical all-MAX_HASH signatures and will be deduplicated against each other (first one kept). That behavior is arguably more correct (empty docs are duplicates of each other), but the description should be corrected to avoid confusion.

tests/ops/deduplicator/test_document_minhash_deduplicator.py
```diff
@@ -1055,3 +1055,37 @@ def test_repeated_process_with_uid_no_pollution(self):
 
 if __name__ == '__main__':
     unittest.main()
+
+
+class DocumentMinhashDeduplicatorEmptyInputTest(DataJuicerTestCaseBase):
```
fengrui-z, line 1056: The new test class is defined after the `if __name__ == '__main__'`: guard (line 1056), so running the file directly with python test_document_minhash_deduplicator.py won't execute these tests (pytest/unittest discovery still finds them). Please move the class above the guard.

tests/ops/deduplicator/test_document_minhash_deduplicator.py
```diff
@@ -1055,3 +1055,37 @@ def test_repeated_process_with_uid_no_pollution(self):
...
+    """Regression test for empty-token ValueError (GitHub issue).
+
+    When a sample's text produces zero tokens (e.g. empty string or text
+    shorter than window_size), compute_hash should not raise ValueError.
+    """
+
+    def test_empty_text_no_crash(self):
```
fengrui-z, line 1064: Tests only cover compute_hash. Consider an end-to-end case: two empty-text samples through `process()` should keep exactly one.

data_juicer/ops/deduplicator/document_minhash_deduplicator.py
cmgzn: c596d16

tests/ops/deduplicator/test_document_minhash_deduplicator.py
cmgzn: c596d16

tests/ops/deduplicator/test_document_minhash_deduplicator.py
cmgzn: c596d16

data_juicer/ops/deduplicator/document_minhash_deduplicator.py
cmgzn: c596d16

fengrui-z (approved):

PR #1052, opened 2026-08-25
fix(analysis): handle pandas StringDtype in overall_analysis

cmgzn:
## Problem

With pandas 2.x, string columns use `StringDtype` instead of `object` dtype. In `OverallAnalysis.refine_single_column()`, the check `if col.dtype != "object"` treats StringDtype columns as numeric, causing downstream `TypeError: Cannot interpret '<StringDtype(na_value=nan)>' as a data type` when numpy operations are applied.

This is the `overall_analysis.py` counterpart to #1047 which fixed the same class of issue in `correlation_analysis.py`.

## Root Cause

`overall_analysis.py` line 48:
```python
if col.dtype != "object":
    return col  # StringDtype passes through here as "not object"
```

Then downstream numpy calls like `np.issubdtype(col.dtype, np.number)` raise TypeError.

## Reproduction

```python
import pandas as pd, numpy as np
col = pd.Series(['zh', 'en', 'fr'], dtype='string')
np.issubdtype(col.dtype, np.number)
# → TypeError: Cannot interpret 'string[python]' as a data type
```

## Fix

Detect `StringDtype` early using `pd.api.types.is_string_dtype()` and route to the string analysis path before the `dtype != "object"` check.

## Tests

Added `OverallAnalysisStringDtypeTest` that creates a dataset with a StringDtype column and verifies `analyze()` completes without error.

---
Tracked in: cmgzn/data-juicer#154

### DIFF
--- a/data_juicer/analysis/overall_analysis.py
+++ b/data_juicer/analysis/overall_analysis.py
@@ -45,8 +45,11 @@ class OverallAnalysis:
         self.supported_object_types = {str, list}
 
     def refine_single_column(self, col):
+        if pd.api.types.is_string_dtype(col) and col.dtype != "object":
+            # pandas 2.x StringDtype: treat like object-dtype strings
+            return col
         if col.dtype != "object":
-            # not an object, return directly
+            # numeric or other non-object dtype, return directly
             return col
         # if the type of this column is object, we can decide the actual type
         # according to the first element.
--- a/tests/analysis/test_overall_analysis.py
+++ b/tests/analysis/test_overall_analysis.py
@@ -102,5 +102,48 @@ class OverallAnalysisTest(DataJuicerTestCaseBase):
         self.assertTrue(os.path.exists(os.path.join(self.temp_output_path, 'overall.md')))
 
 
+class OverallAnalysisStringDtypeTest(DataJuicerTestCaseBase):
+    """Regression test: pandas 2.x StringDtype should not crash analyze()."""
+
+    def setUp(self):
+        super().setUp()
+        self.temp_output_path = 'tmp/test_overall_stringdtype/'
+
+    def tearDown(self):
+        if os.path.exists(self.temp_output_path):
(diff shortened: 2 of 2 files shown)


tests/analysis/test_overall_analysis.py
```diff
@@ -102,5 +102,48 @@ def test_invalid_dataset(self):
...
+        df = pd.DataFrame({
+            Fields.stats: [
+                {'score': 0.8, 'lang': 'zh'},
+                {'score': 0.9, 'lang': 'en'},
+                {'score': 0.7, 'lang': 'fr'},
+            ]
+        })
+        # Force string column to use pandas 2.x StringDtype
+        dataset = NestedDataset.from_pandas(df)
+        flat = dataset.flatten().to_pandas()
```
fengrui-z, line 127: The StringDtype does not survive the dataset round-trip: from_pandas → Arrow string → dataset[Fields.stats] → pd.DataFrame(...) produces an object dtype column, so this test never exercises the new branch (passes without the fix). To make it meaningful, call refine_single_column directly on a StringDtype pd.Series. Also the rename(columns={...}) at line 132 maps columns to their own names (no-op) — please remove.

data_juicer/analysis/overall_analysis.py
```diff
@@ -45,8 +45,11 @@ def __init__(self, dataset, output_path):
         self.supported_object_types = {str, list}
 
     def refine_single_column(self, col):
+        if pd.api.types.is_string_dtype(col) and col.dtype != "object":
+            # pandas 2.x StringDtype: treat like object-dtype strings
+            return col
```
fengrui-z, line 50: Please provide a concrete reproduction (pandas version + code) that fails on current main without this fix. overall_analysis.py has no np.issubdtype call, and describe(percentiles=..., include='all') handles StringDtype fine on pandas 2.3.3. The linked issue references _convert_str_to_number, which doesn't exist in this file.

PR #1053, opened 2026-08-25
fix(exporter): clear error message when export_path has no file extension

cmgzn:
## Problem

When `export_path` has no file extension (e.g. `s3://bucket/data/result`), `Exporter._get_suffix()` returns the entire last path segment as a "suffix", leading to a confusing `NotImplementedError`:

```
NotImplementedError: Suffix of export path [s3://bucket/data/result] or specified
export_type [None] is not supported for now. Only support ['jsonl', 'json', 'parquet'].
```

The message implies the path format is wrong when the actual issue is simply a missing extension.

## Root Cause

`_get_suffix()` uses `export_path.split('.')[-1]` which returns the whole last segment when no dot is present:
```python
's3://bucket/data/result'.split('.')[-1]  # → 's3://bucket/data/result'
```

## Fix

Detect the no-extension case explicitly (using `rsplit('.', 1)` and checking if `/` appears in the "suffix" part) and raise a `ValueError` with actionable guidance:

```
ValueError: Cannot determine file format from export_path [s3://bucket/data/result]
because it has no file extension. Please add a suffix (e.g. .jsonl, .json, .parquet)
or specify export_type explicitly.
```

Also handles edge cases like `/path/to/v1.2/output` where the dot belongs to a directory.

## Tests

Added `ExporterNoSuffixTest` with 4 cases:
- S3 path without extension → ValueError
- Local path without extension → ValueError
- Valid suffix → works normally
- No suffix but explicit `export_type` → bypasses check

---
Tracked in: cmgzn/data-juicer#155

### DIFF
--- a/data_juicer/core/exporter.py
+++ b/data_juicer/core/exporter.py
@@ -179,7 +179,15 @@ class Exporter:
         :param export_path: the path to export datasets.
         :return: the suffix of export_path.
         """
-        suffix = export_path.split(".")[-1].lower()
+        parts = export_path.rsplit(".", 1)
+        if len(parts) < 2 or "/" in parts[-1]:
+            raise ValueError(
+                f"Cannot determine file format from export_path "
+                f"[{export_path}] because it has no file extension. "
+                f"Please add a suffix (e.g. .jsonl, .json, .parquet) "
+                f"or specify export_type explicitly."
+            )
+        suffix = parts[-1].lower()
         return suffix
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


data_juicer/core/exporter.py
```diff
@@ -179,7 +179,15 @@ def _get_suffix(self, export_path):
         :param export_path: the path to export datasets.
         :return: the suffix of export_path.
         """
-        suffix = export_path.split(".")[-1].lower()
+        parts = export_path.rsplit(".", 1)
```
fengrui-z, line 182: Exception type changes from NotImplementedError to ValueError here — verified nothing upstream catches the old type, so this is fine.

fengrui-z (approved):

fengrui-z (approved):

PR #1054, opened 2026-08-25
Add Throughput-Aware Automatic GPU Actor Planning to `ray_partitioned`.

Qirui-jiao:
## Overview

This PR adds a GPU preflight phase before formal `ray_partitioned` execution and automatically assigns different Actor counts to GPU operators based on measured memory usage, steady-state throughput, and output ratios.

The design separates three previously coupled concepts:

- **Logical partition count**: selected by the user or the existing automatic partitioning policy; it controls only checkpoint granularity and recovery boundaries.
- **Actors per operator**: planned globally from measured throughput and cluster resources; the count does not grow directly with the number of partitions.
- **Execution Group**: multiple logical partitions can reuse one Actor Pool, reducing repeated model initialization.

Users can therefore fix the partition count while the framework independently handles operator parallelism and GPU packing.

## Motivation

GPU operators can differ substantially in both throughput and memory demand. A uniform `num_proc` leaves slow operators backlogged and fast operators idle. Using the partition count as an indirect concurrency control also couples checkpoint requirements to compute resources and repeatedly initializes models.

This PR changes the execution flow to:

```text
Load Dataset
  → GPU preflight: measure memory, initialization time, throughput, and output ratio
  → Inject memory / num_gpus
  → Plan a global Actor count for each GPU stage
  → Plan Execution Groups
  → Start the formal Ray pipeline
```

If preflight fails, or the minimum resource plan is infeasible, the formal job is not started.

### 1. Dependency-safe parallel GPU preflight

- Builds a conservative data-dependency graph from each Operator's `input_columns` and `output_columns`.
- Probes GPU targets in parallel when they can be proven independent and depend only on compatible CPU Mappers/Filters.
- Assigns one full GPU to every disposable Ray worker and replays the required CPU ancestors inside that worker.
- With `max_concurrent_gpu_probes: auto`, runs probes serially by default; a positive integer opts into concurrent model loading.
- Safely falls back to ordered recipe replay when column contracts are missing, a GPU data dependency exists, runtime environments are incompatible, or a Dataset-level operation is involved.
- No longer exposes a diagnostic environment marker to custom operators during probing. This prevents extra CUDA synchronization or device-transfer branches and keeps the measured path consistent with the formal Actor path.
- Reports initialization, warmup, steady-state, dependency-replay, and memory-measurement timings from workers, with completion events and heartbeats emitted by the Driver.

Measured memory receives 10% headroom. When resources are not explicitly configured:

```text
planned_memory_mb = ceil(measured_memory_mb × 1.10)
memory_fraction   = ceil(planned_memory_mb / gpu_total_mb × 100) / 100
num_gpus          = max(memory_fraction, 1 / max_gpu_workers_per_device)
```

The current `max_gpu_workers_per_device=5` gives auto-probed operators a Ray scheduling floor of `0.2 GPU`. Operators whose measured memory demand is higher use the actual memory fraction instead.

### 2. Throughput-aware global Actor planning

The planner first assigns one Actor to every eligible GPU stage, then repeatedly scales the current pipeline bottleneck so that source-equivalent throughput is as balanced as possible. Output ratios from filtering stages are included when estimating downstream input volume.

Every additional Actor must satisfy all of the following constraints:

- Cluster CPU budget.
- Ray GPU scheduling fractions.
- Measured per-device GPU memory fractions.
- Maximum model replicas per GPU.
- The number of useful batches available to the stage.
- Capacity reserved for explicitly configured resources and Actor budgets.

Logical partition count is deliberately excluded from Actor planning. If the cluster cannot host the minimum plan of one Actor per eligible stage, the job fails before formal execution.

### 3. Decoupled partition and Actor lifecycles

Actor counts are global job budgets and are no longer copied per partition. Multiple logical partitions can form an Execution Group and share one Actor Pool/model lifecycle:

```text
P logical partitions
  → ceil(P / execution_group_size) Execution Groups
  → initialize the Actor Pool once per group
  → retain checkpoints for every logical partition
```

With `execution_group_size: auto`, preflight initialization time and `max_initialization_overhead_ratio` determine how many partitions should share one Actor lifecycle. A checkpoint resume that cannot preserve grouping safely falls back to per-partition execution.

### 4. Caching, observability, and fault handling

- `gpu_probe_results.json`: memory, throughput, output ratios, dependencies, probe mode, and phase timings.
- Resolved Actor counts and Execution Groups: retained in runtime configuration and logs.
- `dag_execution_plan.json`: physical execution plan.
- JSONL events: probe progress, partition completion, and job completion.
- A probe report is reusable only when the Operator configuration, report version, headroom, profiling parameters, GPU model, and GPU memory capacity still match.
- Ray Actors may be recreated after an unexpected exit; result completeness is covered by integration testing.

## Configuration Example

```yaml
auto_op_parallelism: true

partition:
  mode: manual
  num_of_partitions: 1
  gpu_preflight_enabled: true
  max_gpu_workers_per_device: 5
  max_concurrent_gpu_probes: auto
  gpu_probe_timeout_seconds: null
  gpu_probe_warmup_batches: 1
  gpu_probe_steady_batches: 3
  execution_group_size: auto
  max_initialization_overhead_ratio: 0.1
```

The framework configuration defaults to `1 warmup + 3 steady` batches. To shorten cold-start evaluation time, the real-model integration run below explicitly used `1 warmup + 1 steady`.

### 1. Real nine-operator pipeline on a single 8×H20 node

Environment and workload:

- One node with 8 NVIDIA H20 GPUs and 64 CPUs.
- 20,000 real images.
- Eight GPU workload stages and one CPU metadata stage, plus resize/cleanup helper stages.
- Fixed `num_of_partitions=1`.
- No manually configured `num_proc` for GPU operators; preflight and the planner determine the values automatically.

Automatically resolved plan:

| GPU stage | Actors | Ray `num_gpus` / Actor |
| --- | ---: | ---: |
| rotation | 1 | 0.20 |
| QualiCLIP | 1 | 0.20 |
| clarity | 5 | 0.20 |
| AIGC detection | 1 | 0.20 |
| aesthetic | 10 | 0.25 |
| OCR | 8 | 0.29 |
| RetinaFace | 3 | 0.20 |
| image anomaly | 3 | 0.20 |

```text
Planned Actors: 32
Peak Ray GPU request: 7.62 / 8
Parallel preflight: 16.99 s
Estimated formal execution: 988.71 s
Executor total: 1005.70 s
End-to-end total, including input preparation, Ray startup, and verification: 1040.99 s
Formal-stage throughput: 20.23 images/s
Output: 20,000 / 20,000 rows
```

The reference manually configured script recorded approximately `915 s`. Its timing scope and cache state were not fully identical, so the current result validates automatic planning correctness and resource safety rather than establishing a strict performance comparison.

### 2. Fixed partition-count matrix: 1 / 2 / 4 / 8

A 2×H20 node processed 64 input rows through two GPU Mappers with different throughputs, using fixed partition counts of 1, 2, 4, and 8:

| Partitions | slow / fast Actors | Execution Groups | Observed Actor initializations | Initialization critical-path share | Checkpoint artifacts | Output rows | Executor |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 3 / 1 | 1 | 4 | 1.20% | 2 | 64 | 68.34 s |
| 2 | 3 / 1 | 1 | 4 | 0.98% | 4 | 64 | 78.65 s |
| 4 | 3 / 1 | 1 | 4 | 0.97% | 8 | 64 | 74.45 s |
| 8 | 3 / 1 | 1 | 4 | 0.96% | 16 | 64 | 72.91 s |

Conclusions:

- The Actor plan remains `slow=3, fast=1` and does not multiply with the partition count.
- Every run creates one Execution Group and initializes only four Actors in its Actor Pool.
- Checkpoint count grows with logical partitions, while the Actor lifecycle does not.
- All runs produce the same 64 rows without OOM, GPU overscheduling, or unexpected failures.
- The model-initialization critical-path share remains within `0.96%–1.20%` and does not grow with the partition count.

### 3. Two-node DLC run: 2×8 H20

Environment:

- Two nodes with 8 NVIDIA H20 GPUs each: 16 GPUs and 200 CPUs in total.
- Both nodes use the same Python/Ray environment and identical implementation hashes.
- Sixteen single-GPU stages and 512 input rows.
- Covers normal execution and automatic reconstruction after one Actor exits unexpectedly.

| Case | Preflight GPU slots | Formal Actor distribution | Actor initializations | Output | Executor |
| --- | ---: | ---: | ---: | ---: | ---: |
| normal | 16 | 8 + 8 | 16 | 512 / 512 | 68.12 s |
| actor restart | 16 | 9 + 8, including one replacement | 17 | 512 / 512 | 53.50 s |

Validation results:

- Preflight covers all 16 GPU slots across both nodes, including GPU IDs `0–7` on each node.
- The 16 formal Actors in the normal case are distributed exactly `8 + 8` across the nodes.
- After one injected Actor failure, replacement initialization completes in approximately `7.30 s`; all 512 rows are produced without missing or duplicate results.
- GPU telemetry reports activity on both nodes, with no OOM or unexpected worker failure.
- The Driver takes `168.39 s` for both cases together. The restart case runs second and benefits from process/file caches, so its shorter duration must not be interpreted as a performance improvement from restarting.

The real cluster contains only H20 GPUs. Cache auditing confirms that matching H20 records can be reused while records representing different hardware are rejected. The H20/A100 heterogeneous-cache isolation contract also passes an offline test. A real heterogeneous-GPU cluster and whole-node failure recovery were not exercised in this DLC run.

## Test Coverage

Targeted tests cover:

- Preflight target selection, dependency DAGs, parallel limits, and ordered fallback.
- Memory measurement, 10% headroom, and the `0.2 GPU` scheduling floor.
- Warmup/steady-state throughput, output ratios, and report invalidation.
- Throughput balancing, CPU/GPU/memory packing, explicit-resource priority, and infeasible minimum plans.
- Execution Groups, checkpoints, resume fallback, and the fixed partition matrix.
- Multi-node GPU distribution, hardware-aware cache isolation, and result completeness after Actor restart.

## Compatibility and Limitations

- Automatic probing supports regular CUDA Operators that fit on one GPU. Multi-GPU models still require explicit resource configuration.
- Preflight uses a small sample. Workloads with highly variable sequence length, image resolution, or video duration should provide more representative samples or use more conservative headroom.
- Parallel model loading can be limited by shared-storage and host-memory bandwidth; `max_concurrent_gpu_probes` can be used to throttle it.
- Explicit `memory`, `num_gpus`, and positive `num_proc` values retain priority.
- The two-node test uses lightweight, controllable GPU stages to validate scheduling, distribution, and recovery semantics; it is not a GPU compute benchmark.
- A real heterogeneous-GPU cluster and whole-node failure recovery remain future hardware validation items.

## Conclusion

This PR separates partition granularity, Actor parallelism, and model lifecycle into independent controls. The latest real-model single-node run, fixed partition matrix, and two-node DLC run show that the framework can perform GPU preflight in parallel, assign different Actor counts from measured throughput, keep the global Actor budget stable as partition counts change, and preserve complete output after a multi-node Actor restart.


data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -902,16 +1290,412 @@ def _configure_operator_parallelism(self, ops: List) -> None:
...
+            for op in ops
+            if id(op) in auto_ids
+            and self._is_cuda_operator(op)
+            and float(getattr(op, "_gpu_rows_per_second", 0) or 0) > 0
+            and op.use_ray_actor()
+        ]
+        if not candidates:
+            self._resolved_throughput_actor_plan = None
+            self._throughput_planned_op_ids = set()
+            return None
```
fengrui-z, line 1402: With a default config (np: 4), config.py:1261-1262 injects `num_proc=np` into every op (because `is_ray_mode()` is False at `init_configs` time — Ray isn't up yet). Ops then fail `use_auto_proc()`, `_auto_parallel_op_ids` is empty, and the planner returns None with zero logging. Preflight still runs and burns GPU time, but every stage just gets uniform `num_proc=4` actors — reproduced locally; np: null fixes it. Please either treat np as data-loading-only for `ray_partitioned`, or warn when a profiled op is excluded due to injected `num_proc`, and add an info log when the planner no-ops while probe records exist.

data_juicer/core/executor/gpu_memory_probe.py
```diff
@@ -0,0 +1,1232 @@
...
+        op._gpu_total_mb = float(record.get("gpu_total_mb", 0))
+        profile = record.get("profile") or {}
+        op._gpu_init_seconds = float(profile.get("initialization", 0) or 0)
+        op._gpu_warmup_seconds = float(profile.get("warmup", 0) or 0)
+        op._gpu_steady_batch_seconds = float(profile.get("steady_batch_seconds", 0) or 0)
+        op._gpu_rows_per_second = float(profile.get("steady_rows_per_second", 0) or 0)
+        op._gpu_output_ratio = float(profile.get("output_ratio", 1) or 0)
+        init_kwargs = getattr(op, "_init_kwargs", None)
+        if isinstance(init_kwargs, dict):
+            init_kwargs.update(memory=op.memory, num_gpus=op.num_gpus)
```
fengrui-z, line 1184: (_apply_record) — For a profiled op with explicit `num_gpus`, `_gpu_memory_fraction` comes from measurement while `num_gpus` stays user-set. If measured fraction > num_gpus, your memory-aware bin-packing (_gpu_actor_requests_fit) diverges from Ray's fraction-only packing → possible device OOM. Suggest effective request = max(num_gpus, memory_fraction) or a warning.

data_juicer/core/executor/gpu_memory_probe.py
```diff
@@ -0,0 +1,1232 @@
...
+        ]
+        if barriers:
+            raise RuntimeError(
+                "GPU preflight cannot replay recipe stages across Pipeline operator(s): "
+                f"{', '.join(barriers)}. Configure downstream GPU operators' memory "
+                "and num_gpus explicitly."
+            )
+
+        sample_count = probe_sample_count(list(pending.values()))
+        rows = list(dataset.get(sample_count))
```
fengrui-z, line 804: Sampling is always the dataset head; ordered datasets bias memory/throughput. Add an offset/shuffle option.

data_juicer/core/executor/gpu_memory_probe.py
```diff
@@ -0,0 +1,1232 @@
...
+        raise RuntimeError(f"GPU preflight operator returned inconsistent batch column lengths: {sorted(lengths)}")
+    row_count = lengths.pop()
+    return [{column: values[index] for column, values in batch.items()} for index in range(row_count)]
+
+
+def _construct_probe_op(spec: Mapping[str, Any]):
+    kwargs = dict(spec.get("init_kwargs") or {})
+    # Keep all work in the disposable Ray worker.  Dataset-level
+    # multiprocessing after CUDA initialization would fork an unsafe child.
+    kwargs.update(num_proc=None, auto_op_parallelism=False, skip_op_error=False)
```
fengrui-z, line 419: and :635

`skip_op_error=False` is forced during probing while formal runs default to True; a few bad rows now kill the job at preflight. Honor the recipe's setting for replay/measurement.

fengrui-z (changes requested):

data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -455,29 +602,47 @@ def _configure_auto_partitioning(self, dataset, ops):
                 logger.warning(f"Could not determine dataset size for partition calculation: {e}")
                 logger.info(f"Using fallback partition count: {self.num_partitions}")
 
+        # Keep a real row-count ceiling even when the optimizer is unavailable.
+        # Ray Dataset.split() accepts counts larger than the dataset and creates
```
cmgzn, line 606: One thing to keep in mind is that `Ray Dataset.split(n)` splits by existing blocks rather than by row boundaries. I reproduced this with Ray 2.55.1: a materialized dataset with 1 block and 10 rows produced `[10, 0, 0, 0]` for `split(4)`, while 2 blocks produced `[5, 5, 0, 0]`. Thus, the row-count ceiling above does not prevent empty logical partitions when the block count is smaller than `num_partitions`, and partition sizes can also be highly skewed.

This may affect execution-group balance and checkpoint/resume behavior.

data_juicer/core/executor/ray_executor_partitioned.py
Qirui-jiao: Fixed.

data_juicer/core/executor/gpu_memory_probe.py
Qirui-jiao: Fixed.

data_juicer/core/executor/gpu_memory_probe.py
Qirui-jiao: Fixed.

data_juicer/core/executor/gpu_memory_probe.py
Qirui-jiao: Fixed.

data_juicer/core/executor/ray_executor_partitioned.py

fengrui-z (approved):

PR #1055, opened 2026-08-26
fix(hpo): read gzip JSONL stats samples

mameikagou:
## Problem

`StatsKeys.get_access_log` treated `.jsonl.gz` datasets as plain JSONL while building its HPO analysis sample, causing a `UnicodeDecodeError` before analysis started.

## Fix

- detect supported JSONL suffixes explicitly;
- read gzip JSONL through `gzip.open(..., "rt")` and write a plain `.tmp.jsonl` sample;
- remove the temporary sample even when analyzer setup or execution fails;
- add regression coverage comparing gzip and plain JSONL access logs.

## Validation

- plain JSONL, gzip JSONL, and Zstandard JSONL sample harnesses pass;
- analyzer-error cleanup harness passes;
- isort, flake8, compileall, and `git diff --check` pass.

### DIFF
--- a/data_juicer/utils/constant.py
+++ b/data_juicer/utils/constant.py
@@ -1,4 +1,5 @@
 import copy
+import gzip
 import inspect
 import io
 import os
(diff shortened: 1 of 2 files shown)


cmgzn (approved):
LGTM! Thanks for your work~

PR #1056, opened 2026-08-27
feat: migrate fork repos to personal maintenance

cmgzn:
Transfer recognize-anything and transformers-stream-generator fork repos from the datajuicer org to a personal fork for long-term maintenance, allowing the org-level forks to be archived.

Changes:
- Update all git URLs for recognize-anything and transformers-stream-generator to point to the new fork location
- No functional changes — same code, same fixes, just a different host

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ WORKDIR /data-juicer
 # install basic dependencies for Data-Juicer
 ENV UV_HTTP_TIMEOUT=600
 RUN uv pip install --upgrade --no-cache-dir setuptools==69.5.1 setuptools_scm -i https://pypi.tuna.tsinghua.edu.cn/simple --system \
-    && uv pip install --no-cache-dir git+https://github.com/cmgzn/recognize-anything.git -i https://pypi.tuna.tsinghua.edu.cn/simple --system
+    && uv pip install --no-cache-dir git+https://github.com/cmgzn/recognize-anything.git@889201b76458513d04afde5d4cfa376aa9e33ebf -i https://pypi.tuna.tsinghua.edu.cn/simple --system
 
 # copy source code and install
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ WORKDIR /data-juicer
 # install basic dependencies for Data-Juicer
 ENV UV_HTTP_TIMEOUT=600
 RUN uv pip install --upgrade --no-cache-dir setuptools==69.5.1 setuptools_scm -i https://pypi.tuna.tsinghua.edu.cn/simple --system \
-    && uv pip install --no-cache-dir git+https://github.com/datajuicer/recognize-anything.git -i https://pypi.tuna.tsinghua.edu.cn/simple --system
+    && uv pip install --no-cache-dir git+https://github.com/cmgzn/recognize-anything.git -i https://pypi.tuna.tsinghua.edu.cn/simple --system
 
 # copy source code and install
 COPY . .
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -50,7 +50,7 @@ WORKDIR /data-juicer
 # install basic dependencies for Data-Juicer
 ENV UV_HTTP_TIMEOUT=600
 RUN uv pip install --upgrade --no-cache-dir setuptools==69.5.1 setuptools_scm -i https://pypi.tuna.tsinghua.edu.cn/simple --system \
-    && uv pip install --no-cache-dir git+https://github.com/datajuicer/recognize-anything.git -i https://pypi.tuna.tsinghua.edu.cn/simple --system
+    && uv pip install --no-cache-dir git+https://github.com/cmgzn/recognize-anything.git -i https://pypi.tuna.tsinghua.edu.cn/simple --system
```
fengrui-z: Moving dependencies from an org-level fork to a personal account is a long-term supply-chain risk (account rename/deletion breaks installs). Since the fork HEADs are verified identical right now, please consider pinning to commit SHAs: git+https://github.com/cmgzn/recognize-anything.git@889201b and git+https://github.com/cmgzn/transformers-stream-generator.git@1b356de. This also fixes the pre-existing non-reproducible-build issue (unpinned git+ URLs track the default branch).

data_juicer/utils/model_utils.py
```diff
@@ -39,15 +39,15 @@
 aes_pred = LazyLoader("aesthetics_predictor", "simple-aesthetics-predictor")
 vllm = LazyLoader("vllm")
 diffusers = LazyLoader("diffusers")
-ram = LazyLoader("ram", "git+https://github.com/datajuicer/recognize-anything.git")
+ram = LazyLoader("ram", "git+https://github.com/cmgzn/recognize-anything.git")
```
fengrui-z: Same pinning suggestion applies here and in the three mapper files, so the pinned SHA stays consistent across all 7 locations.

Dockerfile
cmgzn: dd492f6

data_juicer/utils/model_utils.py
cmgzn: dd492f6

Dockerfile
cmgzn: dd492f6

data_juicer/utils/lazy_loader.py
```diff
@@ -252,8 +252,9 @@ def __init__(
...
 
         # Standardize package_url to use git+ format
-        if package_url and "@" in package_url:
-            # Convert from package@git+ format to git+ format
+        if package_url and "@" in package_url.split("git+")[0]:
+            # Convert from package@git+ format to git+ format. Only split on the
+            # 'name @ url' separator, never on a 'url.git@ref' revision pin.
             self._package_url = package_url.split("@", 1)[1]
         else:
             self._package_url = package_url
```
fengrui-z: The new parsing handles `git+https://...git@sha `and `pkg @ git+https://...git@sha`, but breaks raw HTTPS URLs, which `_install_package()` also supports:
```python
LazyLoader(
    "x",
    package_url="https://github.com/org/repo.git@abc",
)
```
In this case, `_package_url` becomes "abc", causing the loader to try installing a package named abc. I confirmed this with a minimal reproduction.
Split only on the explicit PEP 508 separator:
```python
if package_url and " @ " in package_url:
    self._package_url = package_url.split(" @ ", 1)[1]
else:
    self._package_url = package_url
```
Tests should cover raw HTTPS, `git+https`, and `name @ git+https `forms.

cmgzn: 3ed9606

yxdyc (approved):

PR #1057, opened 2026-08-27
fix(ops): correct _supported_exec_modes for deduplicator operators

cmgzn:
## Problem

The `Deduplicator` base class declared `_supported_exec_modes = ("default", "ray", "ray_partitioned")`, but its `run()` method uses HuggingFace Dataset APIs (`NestedDataset`, `dataset.map(num_proc=..., with_rank=..., desc=...)`) which are incompatible with `ray.data.Dataset`. This means regular dedup operators (`ImageDeduplicator`, `VideoDeduplicator`, `DocumentMinhashDeduplicator`, etc.) would crash at runtime if configured in Ray mode, yet the preflight check would not catch it.

Conversely, Ray-only dedup operators (`RayBasicDeduplicator` and its subclasses, `RayBTSMinhashDeduplicator`, `RayBTSMinhashCppDeduplicator`) inherited `"default"` mode from their base classes but cannot run outside Ray.

## Fix

- `Deduplicator` base class: `("default", "ray", "ray_partitioned")` → `("default",)`
- `RayBasicDeduplicator`: add `_supported_exec_modes = ("ray", "ray_partitioned")` (inherited by `RayImageDeduplicator`, `RayVideoDeduplicator`, `RayDocumentDeduplicator`)
- `RayBTSMinhashDeduplicator`: add `_supported_exec_modes = ("ray", "ray_partitioned")`
- `RayBTSMinhashCppDeduplicator`: add `_supported_exec_modes = ("ray", "ray_partitioned")`

Now each operator only declares the execution modes it actually supports, and preflight/runtime checks will correctly reject unsupported configurations.

### DIFF
--- a/data_juicer/ops/base_op.py
+++ b/data_juicer/ops/base_op.py
@@ -920,7 +920,7 @@ class Filter(OP):
 
 
 class Deduplicator(OP):
-    _supported_exec_modes = ("default", "ray", "ray_partitioned")
+    _supported_exec_modes = ("default",)
 
     def __init__(self, *args, **kwargs):
         """
--- a/data_juicer/ops/deduplicator/ray_basic_deduplicator.py
+++ b/data_juicer/ops/deduplicator/ray_basic_deduplicator.py
@@ -107,6 +107,7 @@ class RayBasicDeduplicator(Filter):
 
     # TODO: Set a more reasonable value
     EMPTY_HASH_VALUE = "EMPTY"
+    _supported_exec_modes = ("ray", "ray_partitioned")
 
     def __init__(
         self,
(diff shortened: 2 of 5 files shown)


data_juicer/ops/base_op.py
```diff
@@ -920,7 +920,7 @@ def run(self, dataset, *, exporter=None, tracer=None, reduce=True):
 
 
 class Deduplicator(OP):
-    _supported_exec_modes = ("default", "ray", "ray_partitioned")
+    _supported_exec_modes = ("default",)
```
fengrui-z, line 923: (optional, minor) Consider adding a test in tests/ops/test_base_op.py (near test_all_op_subclasses_declare_supported_exec_modes, line 866) that asserts the expected modes per dedup class, not just that modes exist, to prevent regressions.

cmgzn: [338dd28](https://github.com/datajuicer/data-juicer/pull/1057/commits/338dd282fe59451767038d6970b0cbfec332e4bf)

cmgzn: 338dd28

cmgzn: 338dd28

fengrui-z (approved):

PR #1059, opened 2026-08-28
fix: include text_key and tokenizer in fused context cache keys

arkrolin:
## Problem

Operators in a fused filter group share one `Fields.context` dict per sample.
`fuse_filter_group()` groups ops by intermediate variable
(`op_fusion.py:19-36`, `ALL_INTER_VARS`) and never by `text_key`, so two ops
reading different columns can land in the same group. Several ops then build a
context cache key that does not identify everything the cached value depends
on, and the second op silently reads the first op's value — reporting a
statistic for a column it never looked at.

**Samples are silently dropped as a result.** The wrong statistic is compared
against the op's threshold, so a sample that should pass the filter fails it.
Measured on unfixed code, all three collisions below flip the keep/drop decision
from kept to dropped:

| Intermediate var | Ops differ only in | Statistic         | Unfused | Fused | keep (unfused → fused) |
| ---------------- | ------------------ | ----------------- | ------- | ----- | ---------------------- |
| `INTER_WORDS`    | `text_key`         | `word_rep_ratio`  | `0.0`   | `1.0` | `True` → `False`       |
| `INTER_LINES`    | `text_key`         | `max_line_length` | `30`    | `1`   | `True` → `False`       |
| `refined_words`  | `tokenization`     | `stopwords_ratio` | `1/3`   | `0.0` | `True` → `False`       |

All three are reproduced through the repo's own `load_ops` + `fuse_operators`.

The `INTER_LINES` key was `f"{InterVars.lines}"` — no discriminator at all. The
`refined_words` keys carried neither `model_key` nor `text_key`: at default
settings `word_repetition_filter` built
`f"{InterVars.refined_words}-" f"True-SPECIAL_CHARS-False-[2]-"` while
`stopwords_filter` and `flagged_words_filter` built
`f"{InterVars.refined_words}-True-SPECIAL_CHARS-"` plus their aug settings, both
evaluating to `__dj__refined_words-True-SPECIAL_CHARS-False-[2]-`. That is where
a tokenizing and a non-tokenizing op collided. The `words` keys already carried
`model_key` and needed only `text_key`.

Preconditions: `op_fusion` must be enabled (it defaults to `False`,
`config.py:626`), two or more ops sharing an intermediate variable must be
consecutive in one filter group, and they must differ in `text_key` or
`tokenization`. Single-column pipelines share the cache correctly — that is the
intended behaviour and is unchanged.

## Fix

Added the missing discriminators to all ten affected keys across seven files:
`words_num_filter`, `perplexity_filter`, `word_repetition_filter` (×2),
`flagged_words_filter` (×2), `stopwords_filter` (×2),
`average_line_length_filter`, `maximum_line_length_filter`.

The deliberate cross-op sharing of `refined_words` between
`flagged_words_filter`, `stopwords_filter` and `word_repetition_filter` is
preserved: at default settings all three still produce the identical key
`__dj__refined_words-None-text-True-SPECIAL_CHARS-False-[2]-` (verified).

Context keys are per-batch temporaries — created at `op_fusion.py:171`, popped
at `:185` and `fused_batch_executor.py:87` — and are never persisted. Dataset
fingerprints derive from op config via `generate_fingerprint`, not from these
strings, so no cache invalidation results.

## Tests

`FusedContextKeyIsolationTest` in `tests/ops/test_op_fusion.py` covers each
collision class. Every test runs real ops through `load_ops` + `fuse_operators`
and asserts an explicit expected statistic **and** the resulting keep/drop
decision, fused against unfused:

- `test_inter_words_ops_differing_in_text_key` — `INTER_WORDS`, differing `text_key`
- `test_inter_lines_ops_differing_in_text_key` — `INTER_LINES`, differing `text_key`
- `test_inter_words_ops_differing_in_tokenizer` — `refined_words`, differing `tokenization`

The unfused reference gives every op its own batch, so it cannot read another
op's context.

Verified these are genuine regression tests by reverting all ten cache-key
f-strings to their pre-fix form: all three fail on the statistic, and the
keep/drop measurements in the table above come from that reverted tree. With the
fix restored, `tests/ops/test_op_fusion.py` passes 23/23, and the filter tests
for every op touched here pass 24/24.

**Three existing tests were changed.** They hard-coded the literal key strings
rather than the ops' behaviour, so they pinned the old format:

- `tests/ops/filter/test_ngram_streaming_frequency.py` — `_word_context()` now
  derives both keys from `op.model_key` and `op.text_key`.
- `tests/ops/filter/test_average_line_length_filter.py`,
  `tests/ops/filter/test_maximum_line_length_filter.py` — expected context key
  now includes `text_key`.

No assertion about op output was weakened or removed.

## Why this wasn't caught

`unit-test-partial.yml` is the only PR-triggered workflow, and its
`--mode partial` selects tests via `find_corresponding_test_file()`
(`unittest_utils.py:254`). Each of the five filter files has its own test file,
so `tests/ops/test_op_fusion.py` is never selected when only filters change.
`_run_equal_config` — which does assert fused == unfused — has a single call
site whose ops all use `text_key: 'text'` and `tokenization: false`, so it
could not surface this either. Putting the regression test in
`test_op_fusion.py` means that file is in the diff and CI will run it.

### DIFF
--- a/tests/ops/test_op_fusion.py
+++ b/tests/ops/test_op_fusion.py
@@ -4,7 +4,7 @@ from data_juicer.core import NestedDataset
 from data_juicer.ops.base_op import Mapper, OP
 from data_juicer.ops.load import load_ops
 from data_juicer.ops.op_fusion import fuse_operators, GeneralFusedOP
-from data_juicer.utils.constant import Fields, InterVars
+from data_juicer.utils.constant import Fields
 from data_juicer.utils.unittest_utils import DataJuicerTestCaseBase
 
 
@@ -2238,7 +2238,9 @@ class FusedContextKeyIsolationTest(DataJuicerTestCaseBase):
     value and reports a statistic for a column it never looked at.
     """
 
-    def _stats_fused_vs_unfused(self, process_list, row):
+    def _fused_vs_unfused(self, process_list, row):
+        """Compute stats and the keep/drop decision, fused and unfused."""
+
         def make_batch():
             batch = {key: [value] for key, value in row.items()}
             batch[Fields.stats] = [{}]
@@ -2252,14 +2254,32 @@ class FusedContextKeyIsolationTest(DataJuicerTestCaseBase):
         fused_batch = make_batch()
         fused_ops[0].compute_stats_batched(fused_batch)
         fused_stats = fused_batch[Fields.stats][0]
+        fused_keep = bool(list(fused_ops[0].process_batched(fused_batch))[0])
 
-        unfused_stats = {}
+        # Every unfused op gets its own batch and therefore its own context,
+        # so it cannot read another op's intermediate value. That makes the
+        # unfused run the reference behaviour the fused run has to match.
+        unfused_stats, unfused_keep = {}, True
         for op in load_ops(process_list):
             batch = make_batch()
             op.compute_stats_batched(batch)
             unfused_stats.update(batch[Fields.stats][0])
+            unfused_keep &= bool(list(op.process_batched(batch))[0])
 
-        return fused_stats, unfused_stats
+        return fused_stats, unfused_stats, fused_keep, unfused_keep
+
+    def _assert_fusion_transparent(self, process_list, row, expected_stats,
+                                   expected_keep):
+        fused_stats, unfused_stats, fused_keep, unfused_keep = \
+            self._fused_vs_unfused(process_list, row)
+
+        self.assertEqual(sorted(unfused_stats), sorted(expected_stats))
+        self.assertEqual(sorted(fused_stats), sorted(expected_stats))
+        for key, expected in expected_stats.items():
+            self.assertAlmostEqual(unfused_stats[key], expected, places=6)
+            self.assertAlmostEqual(fused_stats[key], expected, places=6)
+        self.assertEqual(unfused_keep, expected_keep)
+        self.assertEqual(fused_keep, expected_keep)
 
     def test_inter_words_ops_differing_in_text_key(self):
         process_list = [{
@@ -2281,14 +2301,19 @@ class FusedContextKeyIsolationTest(DataJuicerTestCaseBase):
             }
         }]
         # 'text' is all-identical 2-grams, 'translation' has no repetition, so
-        # reading the wrong column flips word_rep_ratio from 0.0 to 1.0.
+        # reading the wrong column flips word_rep_ratio from 0.0 to 1.0, which
+        # also pushes the sample past max_ratio and drops it.
         row = {
             'text': 'ha ha ha ha ha ha ha ha',
             'translation': 'alpha beta gamma delta epsilon zeta eta theta'
         }
-        fused_stats, unfused_stats = self._stats_fused_vs_unfused(
-            process_list, row)
-        self.assertEqual(fused_stats, unfused_stats)
+        self._assert_fusion_transparent(process_list,
+                                        row,
+                                        expected_stats={
+                                            'num_words': 8,
+                                            'word_rep_ratio': 0.0
+                                        },
+                                        expected_keep=True)
 
     def test_inter_lines_ops_differing_in_text_key(self):
         process_list = [{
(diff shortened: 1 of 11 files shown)


tests/ops/test_op_fusion.py
```diff
@@ -2227,5 +2227,138 @@ def test_inplace_then_newdict_mapper(self):
...
+            'words_num_filter': {
+                'lang': 'en',
+                'tokenization': False,
+                'text_key': 'translation'
+            }
+        }])
+        keys = [f'{InterVars.words}-{op.model_key}-{op.text_key}'
+                for op in ops]
+        self.assertEqual(len(set(keys)), 2,
+                         'ops reading different columns must not share a key')
```
cmgzn: This assertion only checks the key format constructed inside the test; it never checks which key `WordsNumFilter` actually uses. It therefore still passes if the production code uses the old key without `text_key`—I confirmed that it passes against unchanged main (`70b7697`). Also, both operators have `tokenization=False`, so this does not exercise tokenizer separation despite the test name.

Could we replace this with an execution-based assertion, or extend the existing `test_inter_words_ops_differing_in_text_key` instead? Comparing actual fused/unfused statistics against an explicit expected value would detect a regression in the production code.

tests/ops/test_op_fusion.py
```diff
@@ -2227,5 +2227,138 @@ def test_inplace_then_newdict_mapper(self):
...
+        def refined_words_key(op):
+            return (f'{InterVars.refined_words}'
+                    f'-{op.model_key}-{op.text_key}'
+                    '-True-SPECIAL_CHARS-'
+                    f'{op.use_words_aug}-'
+                    f'{op.words_aug_group_sizes}-'
+                    f'{op.words_aug_join_char}')
+
+        self.assertNotEqual(refined_words_key(untokenized),
+                            refined_words_key(tokenized))
```
cmgzn: This test compares the output of its own `refined_words_key()` helper after changing an attribute; neither operator computes any statistics, and both were constructed with `tokenization=False`. It still passes against unchanged main, so the tokenizer collision currently has no behavioral regression coverage here.

Could we exercise a tokenized `word_repetition_filter` followed by an untokenized `flagged_words_filter` on the same field through `load_ops` + `fuse_operators`? With `hello alpha beta gamma delta epsilon zeta theta` and a local flagged-word list containing only `hello`, I reproduced `flagged_words_ratio=0.125` unfused versus `0.0` fused on main using the real English SentencePiece model; the results agree with this fix. Please assert the expected statistic and fused/unfused agreement (ideally also keep/drop behavior), and verify the test fails with the production fix reverted.

cmgzn (commented):
Two inline comments on the cache-isolation tests. I ran this PR's four new tests against unchanged main (`70b7697`): the two behavioral tests fail as expected, but both key-construction tests still pass.

tests/ops/test_op_fusion.py
arkrolin: Thanks for the careful review — I'm sorry the original test setup was too weak.

Both tests you flagged asserted on key strings the tests themselves constructed,
so neither ever observed which key the production code uses. Both ops in both
tests used `tokenization=False`, so there was no tokenizer difference to detect
either. They passed against unchanged main exactly as you found, so the
tokenizer collision had no behavioral coverage at all. So I replaced them.

## Changed

`FusedContextKeyIsolationTest` now holds three execution-based tests. Each runs
real ops through `load_ops` + `fuse_operators` and asserts an explicit expected
statistic and the keep/drop decision, fused against unfused:

| Test | Intermediate var | Ops differ in |
| --- | --- | --- |
| `test_inter_words_ops_differing_in_text_key` | `INTER_WORDS` | `text_key` |
| `test_inter_lines_ops_differing_in_text_key` | `INTER_LINES` | `text_key` |
| `test_inter_words_ops_differing_in_tokenizer` | `refined_words` | `tokenization` |

The shared helper asserts the stat key set, each expected value, and the
keep/drop result for both paths. The unfused reference gives every op its own
batch, so it cannot read another op's context.

```python
# tokenizer case
row = {'text': 'the quick brown fox jumps over the lazy dog'}
self._assert_fusion_transparent(process_list,
                                row,
                                expected_stats={
                                    'word_rep_ratio': 0.0,
                                    'stopwords_ratio': 1 / 3
                                },
                                expected_keep=True)
```

## Verification with the production fix reverted

I reverted all 10 fixed cache-key f-strings across the 7 source files to their
pre-fix form and re-ran the three tests. All three fail:

```
E   AssertionError: 1 != 30 within 6 places (29 difference)                    # max_line_length
E   AssertionError: 1.0 != 0.0 within 6 places (1.0 difference)                # word_rep_ratio
E   AssertionError: 0.0 != 0.3333333333333333 within 6 places (0.33... diff)   # stopwords_ratio

FAILED tests/ops/test_op_fusion.py::FusedContextKeyIsolationTest::test_inter_lines_ops_differing_in_text_key
FAILED tests/ops/test_op_fusion.py::FusedContextKeyIsolationTest::test_inter_words_ops_differing_in_text_key
FAILED tests/ops/test_op_fusion.py::FusedContextKeyIsolationTest::test_inter_words_ops_differing_in_tokenizer
============================== 3 failed in 3.72s ===============================
```

The statistic assertion trips first, so I also measured the keep/drop effect
directly on the reverted tree. Every sample flips from kept to dropped:

```
T1 words / text_key
   fused  : {'num_words': 8, 'word_rep_ratio': 1.0}                keep= False
   unfused: {'num_words': 8, 'word_rep_ratio': 0.0}                keep= True
T2 lines / text_key
   fused  : {'avg_line_length': 1.666..., 'max_line_length': 1}    keep= False
   unfused: {'avg_line_length': 1.666..., 'max_line_length': 30}   keep= True
T3 words / tokenizer
   fused  : {'word_rep_ratio': 0.0, 'stopwords_ratio': 0.0}        keep= False
   unfused: {'word_rep_ratio': 0.0, 'stopwords_ratio': 0.333...}   keep= True
```

So the collision is not cosmetic: it silently drops samples that should be kept.

With the fix restored, `tests/ops/test_op_fusion.py` passes 23/23, and the filter
tests for every op touched here pass 24/24.

## Two deviations worth flagging

**I used `stopwords_filter`, not `flagged_words_filter`, as the second op in the
tokenizer test.** `FlaggedWordFilter` takes `flagged_words_dir` and loads its
list from the JSON assets via `load_words_asset` — there is no
`flagged_words_file` parameter — so pointing it at a local single-word list
inside a unit test isn't possible without writing into `ASSET_DIR` or patching
the loader. `stopwords_filter` reproduces the same collision on the same code
path with no fixture setup, and its stat key differs from
`word_repetition_filter`'s, which the assertion needs.

**Your reproduction and mine differ in which key collides.** The `words` keys
already carried `model_key` before this PR; the omission was in the
`refined_words` keys. At default settings `word_repetition_filter` built
`f"{InterVars.refined_words}-" f"True-SPECIAL_CHARS-False-[2]-"` while
`stopwords_filter` and `flagged_words_filter` built
`f"{InterVars.refined_words}-True-SPECIAL_CHARS-"` plus their aug settings —
both evaluating to `__dj__refined_words-True-SPECIAL_CHARS-False-[2]-`. Same
bug, same fix, but the collision is on the `refined_words` path rather than
`words`.

Sorry that the original PR description was wrong on this point — it claimed the
omission was `model_key` in `words_num_filter`. I've rewritten the whole
description to match what the code actually does, including the ten affected
keys and the keep/drop impact above. Thanks again for the review.

cmgzn (approved):
LGTM. Thanks for your work~

PR #1048, opened 2026-08-20
docs: rewrite guides, add new operator docs, and fix code examples

cmgzn:
## Summary

Refresh the English and Chinese documentation, correct examples and configuration references, and organize the Sphinx site around separate Doc and API sections. Update deployment to build the affected version while retaining previously published versions.

## Documentation

- Add bilingual guides for data processing, analysis, global configuration, and the playground; restructure Quick Start, installation, dataset configuration, and export documentation.
- Add documentation for 28 operators covering agent/dialog quality, PII, video/robotics, and supporting utilities, and refresh the operator catalog.
- Correct API names, configuration defaults, Python requirements, and cross-page links. Standardize Chinese-only filenames with `_ZH` and update references in demos, tools, and shell help.

## Navigation

- Reorganize the English and Chinese landing pages and documentation indices into task-oriented guide groups, with a separate API entry point.
- Integrate the theme's top-level Doc/API tabs and section-specific sidebars, implemented in [data-juicer-sphinx#17](https://github.com/datajuicer/data-juicer-sphinx/pull/17).
- Historical-version source preservation is handled separately in [data-juicer-sphinx#18](https://github.com/datajuicer/data-juicer-sphinx/pull/18), so older tags retain their own documentation indices.

## Build and supporting fixes

- Build PR previews, main, or the pushed tag as appropriate; provide an explicit full-rebuild option and preserve existing published versions during incremental deployment.
- Declare documentation dependencies explicitly, use CPU PyTorch packages, and load the theme extension. Reuse the theme's shipped Ask-AI bundle, removing redundant Node/npm build steps.
- Resolve operator recipe names from registration syntax without importing operators, and defer translator imports until translation is needed.
- Fix package imports in the arXiv downloader and HPO entry point; put W&B sweep execution behind `main()` so importing the module does not start a sweep. Add regression coverage for operator metadata and HPO behavior.

## Validation

- Local English and Chinese previews covered main, this branch, and tags v1.5.0–v1.5.5, including Doc/API navigation and responsive layouts.
- Preview builds used `-A` to skip API rendering; these checks do not establish complete API coverage. Historical documentation warnings remain.
- With the companion theme fix, the actual `--tags v1.5.0 -A` build path succeeded in both languages and excluded a deliberately added current-version guide from the historical output.
- Workflow YAML, shell syntax, and the latest commit's pre-commit checks passed.

PR #686, opened 2025-06-04
Refactor sandbox

HYLcool:
- refactor sandbox
- support multiple pipelines
- support context infos
- support env manager to run different commands in different envs.
- add sandbox recipes for InternVL COCO Caption task

PR #1061, opened 2026-08-31
fix: align global configuration with runtime behavior

cmgzn:
Several global configuration fields are accepted without affecting the built-in executors, while existing runtime controls are missing from YAML/CLI. Reader options also reach different execution and analysis paths inconsistently. This PR aligns those declarations, examples, and consumers with the implementation.

This is an independent follow-up to the configuration audit discussed during #1048. Partition-size compatibility, sample targets, and `ModalityConfig` cleanup are intentionally handled by #1063 and are excluded here.

**Changes**

- Remove 12 unused declarations: all eight `intermediate_storage.*` fields, legacy `preserve_intermediate_data`, `resource_optimization.auto_configure`, `max_log_size_mb`, and `backup_count`.
- Drop the stale `add_suffix` key from `config_min.yaml`. #686 turned it into a value derived from the presence of a `suffix_filter` operator, computed in `load_strategy.py`; `config_all.yaml` dropped it at the time, but `config_min.yaml` kept it and `get_default_cfg()` reads that file with plain `yaml.safe_load()`, so the orphan never failed validation. The file is now fully parseable by `init_configs`.
- Expose the existing `load_jsonl_lenient` and `use_dag` controls, preserving executor defaults.
- Apply global reader defaults in `DatasetBuilder`, with explicit call arguments taking precedence. `read_options` and `override_num_blocks` were read from `self.cfg` inside `RayExecutor.run()` and `RayExecutorPartitioned._run_impl()` and passed as explicit arguments; they now come from the builder, so `RayAnalyzer` -- which called a bare `load_dataset()` -- gets the same reader behavior as execution, and callers can still override. Forward Ray local JSON `read_options`, convert YAML dictionaries to PyArrow `ReadOptions`, and preserve reader arguments in the older-PyArrow fallback.
- Make `np` the default for loader parallelism and deprecate the `load_data_np` run argument (see below).
- Make partition analysis use the first configured `text_keys` field, including nested paths and an empty list, while retaining the programmatic `text_key` fallback.
- Reject the unsupported `checkpoint.strategy=every_partition` value and nonpositive checkpoint intervals and read block counts.
- Replace the `notification` and `annotation` example blocks in `config_all.yaml` with a pointer to the per-operator settings that actually consume them (`notification_config` / `notification_events`). Neither block was ever a global declaration.
- Update help text, global examples, partition/checkpoint demos, and English/Chinese documentation.

**Loader parallelism: `np`, `load_dataset_kwargs.num_proc`, and `load_data_np`**

Three code paths could set the loader's `num_proc`, and they disagreed about which one wins.

`load_dataset_kwargs` was applied *after* the explicit call arguments, so a `num_proc` sitting in the recipe outranked whatever the caller passed — including the value the executors injected from `np`. Applying the recipe defaults first, as the reader-option change does, fixes that inversion but makes the recipe key dead instead: the `Default*` load strategies read `kwargs.get("num_proc", 1)`, falling back to a single process rather than to `np`, so once the executor always injected a value nothing else could take effect.

This PR seeds `num_proc` from `np` in `DatasetBuilder`, after both sources and only when neither supplied a value. Precedence now reads, weakest to strongest:

1. `np` — the global default, applied for the default executor and the Analyzer.
2. `load_dataset_kwargs.num_proc` — a recipe-level override, for decoupling load parallelism from processing parallelism.
3. An explicit `DatasetBuilder.load_dataset(num_proc=...)` argument.

`num_proc` was never part of #922's intent for `load_dataset_kwargs` (that PR names chunksize, columns, and delimiter); it worked only as a side effect of the ordering. Layer 2 is now the intended way to express the decoupling, and it is documented as such.

That leaves `load_data_np`, which resolved to the same `num_proc` keyword through the same channel, but had a narrower reach than its signature suggested: it is declared on every `run()`, yet only the default executor and the Analyzer applied the `self.np` fallback and forwarded it. No `Ray*` load strategy reads `num_proc` — they select keys explicitly and never splat `**kwargs` — and `RayAnalyzer` accepted the argument and dropped it outright. It is also reachable only from Python; nothing in the repository, the configuration schema, or the documentation passes it.

Since `load_data_np` is the first positional parameter of `RayExecutor.run()` and `RayExecutorPartitioned.run()`, removing it outright would break positional callers silently. The parameter is kept and routed through a shim that logs a deprecation warning and hands the value to the same `num_proc` channel, so existing calls keep their behavior.

The shim receives the builder's executor type, because one warning text cannot describe both outcomes without misleading half of its readers. On the default executor and the Analyzer the value is applied, so the warning names the parallelism in use before pointing at `load_dataset_kwargs.num_proc`; a bare deprecation notice there would read as though the argument had been dropped. On the Ray executors it says the argument has never been applied and points at `override_num_blocks`. The Ray `run()` docstrings previously named `load_dataset_kwargs.num_proc` as the replacement, which is ignored on those paths for exactly the same reason, and are corrected. The `self.np` fallbacks are gone from the executors — that job now belongs to `DatasetBuilder` alone, which is also what gives the Ray executors a defined `np` default for the first time.

**Documentation**

Reader options and loader parallelism are documented in `docs/DatasetCfg.md`, `use_dag` in `docs/JobManagement.md`, and the checkpoint strategy constraints in `docs/PartitionAndCheckpoint.md`, each with its Chinese counterpart. An earlier revision of this PR added standalone `docs/ConfigMigration.md` / `_ZH.md` files; those were folded into the feature documents above, since a migration guide dates while the feature docs stay current. The removed-key list below is the part with no evergreen home and belongs in the release notes:

| Removed key | Notes |
| --- | --- |
| `intermediate_storage.preserve_intermediate_data` | Never read by any executor. |
| `intermediate_storage.cleanup_temp_files` | Never read by any executor. |
| `intermediate_storage.cleanup_on_success` | Never read by any executor. |
| `intermediate_storage.retention_policy` | Never read by any executor. |
| `intermediate_storage.max_retention_days` | Never read by any executor. |
| `intermediate_storage.format` | Never read by any executor. |
| `intermediate_storage.compression` | Never read by any executor. |
| `intermediate_storage.write_partitions` | Never read by any executor. |
| `preserve_intermediate_data` | Superseded by checkpoint management. |
| `resource_optimization.auto_configure` | Partition sizing is always computed. |
| `max_log_size_mb` | Log rotation is handled by loguru defaults. |
| `backup_count` | Log rotation is handled by loguru defaults. |

**Compatibility**

Recipes containing the 12 removed keys now fail configuration validation and must delete those keys. `every_partition` previously fell back to `every_op`; recipes should select a supported strategy explicitly. `run(load_data_np=...)` still works and now warns. Recipes that relied on `load_dataset_kwargs.num_proc` keep working, and it now overrides `np` rather than overriding the caller. Existing checkpoint creation, resumption, and temporary-directory cleanup remain unchanged.

Sandbox probe settings, `hpo_config`, and the `partition_dir` integration interface remain available. The audit covered 97 original declarations and references across 505 non-test Python files; static reference counts were only a starting point for checking runtime consumers.

**Validation**

All Python validation used `uv run` with Python 3.11.

- 163 distinct targeted tests passed across configuration behavior, `DatasetBuilder`, load strategies, Ray datasets, optimizer text-key analysis, and the complete `PartitionedRayExecutor` test module.
- The executor slice includes real local-Ray execution and checkpoint-disabled behavior.
- New tests cover the three-layer `num_proc` precedence and the `load_data_np` shim: that it stays silent when unset, and that each executor type gets the warning describing what actually happened.
- A trial merge of the latest #1061 and #1063 heads completed without conflicts; their combined focused suite passed all 165 tests.
- Commit hooks passed, including isort, Black, flake8, YAML checks, operator documentation generation, and secret detection.
- Earlier regression checks reproduced the missing-control, ignored-reader-option, and ignored-`text_keys` behavior before the fixes. Mutations affecting reader defaults, DAG selection, Ray options, and checkpoint selection were detected by the tests.

**Remaining validation limits**

The full pytest suite previously stopped during collection because `tests/tools/test_mcp_server.py` requires the unavailable `mcp` package. Remote S3/HDFS, GPU operators, and historical PyArrow versions were not exercised. The older-PyArrow fallback was reviewed in source; runtime checks used the installed version. `tests/core/data/test_dataset_error_propagation.py` has three failures that reproduce identically on the base commit, and `tests/core/test_analyzer.py::test_auto_analysis_with_existing_dataset` waits on a network fetch in the local environment; neither touches the changed paths. This remains a draft for further review.


fengrui-z (approved):
LGTM

PR #1062, opened 2026-08-31
feat: add litellm provider

prodmanpd:
## Summary
- Adds an opt-in LiteLLM backend in data_juicer/utils/model_utils.py. Pass use_litellm=True to prepare_api_model(...) and chat/embedding calls route through litellm.completion / litellm.embedding.
- One provider/model string reaches 100+ providers via LiteLLM (https://github.com/BerriAI/litellm), including ones the OPENAI_BASE_URL override can't reach because their auth isn't OpenAI-compatible: Bedrock (SigV4), Vertex (service accounts), Azure AD.
- Fully additive and off by default: use_litellm=False keeps the existing path unchanged.

## Changes
- `data_juicer/utils/model_utils.py`
  - `litellm = LazyLoader("litellm")` (mirrors `openai = LazyLoader("openai")`; no import cost unless used).
  - `LiteLLMChatAPIModel`, `LiteLLMEmbeddingAPIModel`, and `LiteLLMResponsesAPIModel` — same `__call__` interface as the OpenAI classes, covering all three endpoints (`/chat/completions`, `/embeddings`, `/responses`). Because LiteLLM returns OpenAI-shaped objects, `response.model_dump()` feeds the **existing** `nested_access(result, response_path)` extraction unchanged (`choices.0.message.content` / `data.0.embedding` / `output.0.content.0.text`).
  - `prepare_api_model(..., use_litellm=False)` selects the LiteLLM classes when enabled (and skips the DashScope base-url model remap, since LiteLLM routes by the model string).
- `tests/utils/test_model_utils.py` — `test_prepare_api_model_litellm` (dispatch, `drop_params` default, credential omission, embedding path, model-required guard).
- `pyproject.toml` — `litellm>=1.89.0,<2` added to the `ai_services`.

## Implementation notes
- **`drop_params=True`** is defaulted at the call site so providers that reject a given kwarg (e.g. Anthropic on `seed`/`frequency_penalty`) don't break the shared path.
- **Credentials omitted when blank** (`_split_litellm_credentials`): if `api_key`/`base_url` aren't set, nothing is passed, so LiteLLM falls back to the target provider's own env var (`ANTHROPIC_API_KEY`, `GEMINI_API_KEY`, AWS creds, ...). 

## Tests

**1. Unit test** — `pytest tests/utils/test_model_utils.py::ModelUtilsTest::test_prepare_api_model_litellm -v`
```
tests/utils/test_model_utils.py::ModelUtilsTest::test_prepare_api_model_litellm PASSED [100%]
========================= 1 passed in 75.10s =========================
```

**2. Lint** — `flake8` (flake8-pyproject, max-line 120) reports **no new issues** on the changed lines.

**3. Live end-to-end** through the real `prepare_api_model(use_litellm=True)` path (LiteLLM proxy), two providers via one code path:
```
chat  (gpt-4.1-mini)     -> 'DJ_LITELLM_OK'   usage: {'prompt_tokens': 17, 'completion_tokens': 6, 'total_tokens': 23}
chat  (gemini-2.5-flash) -> '4'
```
This exercises `prepare_api_model` -> `LiteLLMChatAPIModel` -> `litellm.completion` -> provider -> `model_dump()` -> `nested_access`.

**4. Pin resolution** — `litellm>=1.89.0,<2` resolves to `1.98.0` cleanly alongside the repo's `openai` (`2.54.0`).

## Example usage
```python
from data_juicer.utils.model_utils import prepare_api_model

# Bedrock via native SigV4 (not reachable through OPENAI_BASE_URL):
model = prepare_api_model(
    "bedrock/anthropic.claude-3-5-sonnet-20241022-v2:0",
    use_litellm=True,
)  # AWS creds picked up from the environment
print(model([{"role": "user", "content": "Rate this text 1-5..."}]))

# Or a self-hosted LiteLLM proxy:
model = prepare_api_model(
    "litellm_proxy/gpt-4.1-mini", use_litellm=True,
    base_url="http://localhost:4000", api_key="sk-...",
)
```

### DIFF
--- a/data_juicer/utils/model_utils.py
+++ b/data_juicer/utils/model_utils.py
@@ -659,6 +659,8 @@ def prepare_api_model(
     endpoint = endpoint or "/chat/completions"
 
     if api_backend == "litellm":
+        # get_model() supplies the local execution device, not an API parameter.
+        model_params.pop("device", None)
         # LiteLLM routes by the ``provider/model`` string and each provider
         # reads its own credentials, so we must NOT merge OpenAI/DashScope env
         # vars into the params here (that would leak one provider's key/base_url
--- a/data_juicer/utils/model_utils.py
+++ b/data_juicer/utils/model_utils.py
@@ -603,7 +603,7 @@ def prepare_api_model(
     response_path=None,
     return_processor=False,
     processor_config=None,
-    use_litellm=False,
+    api_backend=None,
     **model_params,
 ):
     """Creates a callable API model for interacting with OpenAI-compatible API.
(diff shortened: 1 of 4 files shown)


cmgzn:
Thanks for the contribution! The opt-in LiteLLM integration looks useful.

Could we replace `use_litellm` with an explicit `api_backend` option?

```python
prepare_api_model(
    ...,
    api_backend="openai_compatible",  # or "litellm"
)
```

For operator configs, it can remain under `model_params`:

```yaml
model_params:
  api_backend: litellm
```

This makes the behavior clearer to users: `litellm` uses the LiteLLM SDK for provider-specific request conversion, while `openai_compatible` uses the existing direct OpenAI-compatible path. The default should remain `openai_compatible`, so existing configurations keep their current behavior.

There is also one ordering issue in the current implementation: `_merge_openai_compatible_env_into_model_params()` runs before the backend is selected. On the LiteLLM path, this can convert `OPENAI_API_KEY` or `OPENAI_BASE_URL` into explicit arguments passed to LiteLLM, even when the selected model uses another provider.

Could we select `api_backend` first and apply the existing OpenAI/DashScope environment-variable merging only to the `openai_compatible` path? Explicit `api_key` and `base_url` values supplied through `model_params` should still be honored by the LiteLLM path.

cmgzn (approved):
LGTM, Thanks for your work

liuzhao1225:
I'm concerned about the supply-chain boundary of this LiteLLM integration. The [upstream incident report](https://docs.litellm.ai/blog/security-update-march-2026) confirms compromised PyPI releases 1.82.7/1.82.8 in March 2026; those releases were removed.

At reviewed head `d72e6263cf`, this PR includes `litellm>=1.80.0,<2` in `pyproject.toml`.

The reviewed lockfile pins 1.80.0, and its LiteLLM artifact URLs/hashes matched PyPI metadata. Lock-respecting installs retain that version; manifest resolution without the lock can differ. Please document the supported install path and how dependency updates are tested and verified.

This is one of [49 observed LiteLLM integration PRs](https://github.com/liuzhao1225/litellm-integration-supply-chain-review/blob/main/PR-INDEX.md) from the same account, whose [author acknowledged the cross-project effort](https://github.com/chartbrew/chartbrew/pull/365#issuecomment-5207856315). The pattern prompted this review; it does not establish malicious intent or connect the author to the incident. The consolidated questions and corrections are in [YouDub #130](https://github.com/liuzhao1225/YouDub-webui/pull/130); this PR's [review snapshot](https://github.com/liuzhao1225/litellm-integration-supply-chain-review/blob/main/prs/datajuicer__data-juicer__1062.json) preserves the revision and scope.

This is a follow-up documentation note on the reviewed historical change; it does not assert that the current release is compromised.

PR #1063, opened 2026-09-01
fix: align partition size configuration

cmgzn:
## Problem

The partition configuration exposed multiple fields whose runtime behavior did not match their names or documentation:

- The legacy flat `max_partition_size_mb` parameter was parsed but never consumed by the current executor or optimizer.
- `partition.target_size_mb` is the active MB input used by the auto-mode optimizer for planning; it is an estimate rather than a hard partition-size limit.
- The legacy flat `partition_size` parameter had become disconnected from execution, while the executor read an undocumented nested `partition.size` value with no parser declaration.
- Manual configuration could otherwise silently combine an explicit partition count with a samples-per-partition target.

## Changes

- Remove the unused flat `max_partition_size_mb` parameter.
- Keep `partition.target_size_mb` as the user-facing MB planning input.
  - Zero and negative values remain valid inputs.
  - When a calculation falls below an optimizer minimum, the optimizer applies its 100-sample, 10-sample, or 32-MB lower bound and logs a warning.
- Add `partition.size` as an optional samples-per-partition target:
  - manual mode derives the nearest partition count after loading the dataset;
  - auto mode uses it when the optimizer cannot return a valid sample-count recommendation.
- For manual `partition.size`, materialize the input once and split at row boundaries:
  - use `size`, `2 * size`, ... as boundaries and let the final partition absorb the nearest-count remainder;
  - return one materialized dataset directly when the derived count is one;
  - keep explicit-count and auto-mode partitioning on their existing block-based path.
- Make `partition.size` and `partition.num_of_partitions` mutually exclusive in manual mode.
- Continue accepting an explicitly configured flat `partition_size` with a deprecation warning:
  - when no manual count is configured, it migrates to `partition.size`;
  - when an explicit manual count exists, the count is preserved and the deprecated field is ignored with a warning;
  - an explicit nested value takes precedence in auto mode.
- Remove executor reads of the nonexistent nested `partition.max_size_mb` setting and label the dynamic MB recommendation as an estimate.
- Reduce `ModalityConfig` to the fields that participate in sample-count calculation:
  - `default_partition_size` supplies a fallback when a sample-size estimate is unavailable;
  - `max_partition_size` caps the calculated sample count;
  - duplicated modality metadata, the stale static MB value, unused multipliers, and description text are removed.
- Record the resolved nested sample target and actual final partition count in the detailed job-start event.
- Update the focused English and Chinese documentation and configuration example.

## Compatibility

Configs containing `max_partition_size_mb` now fail validation instead of being silently accepted and ignored. Auto-mode MB planning uses:

```yaml
partition:
  target_size_mb: 256
```

Manual sample-target and explicit-count forms are alternatives:

```yaml
partition:
  mode: manual
  size: 5000
```

```yaml
partition:
  mode: manual
  num_of_partitions: 8
```

The old flat `partition_size` remains accepted for migration and emits a `DeprecationWarning`.

`get_partition_recommendations()["modality_configs"]` remains available, but each modality now reports only `default_size` and `max_size`. The removed metadata did not participate in optimizer calculations. The top-level `recommended_max_size_mb` result remains as a dynamic estimate for compatibility.

## Validation

- Focused config, auto-partition, and manual-size edge tests: 48 passed
- Optimizer and event-logging tests: 79 passed
- Full `PartitionedRayExecutor` suite: 29 passed
- Real local-Ray E2E with 1,001 rows and `size: 500`:
  - nested manual config: two partitions with `500 / 501` rows
  - deprecated flat config: the same result with one deprecation warning
  - forced auto-optimizer failure remained on the existing auto-mode split path
  - all cases preserved 1,001 output rows and logged `partition_size=500`, `total_partitions=2`
- Real optimizer runs with `target_size_mb` set to `0` and `-64` both produced the optimizer minimum recommendations (`100` samples and `32` MB) and emitted explicit clamp warnings
- Mutation checks confirmed that the new tests fail if nearest-count calculation, manual conflict validation, or clamp warnings are removed
- Deterministic before/after comparisons for text, image, audio, video, and multimodal inputs produced identical partition-count and dynamic-MB recommendations after the `ModalityConfig` cleanup
- A mutation check confirmed that recommendation-schema tests reject reintroduced stale modality metadata
- Single-block E2E: 11 rows with `size: 5` produced `5 / 6`; 11 rows with `size: 10` produced one 11-row partition
- Mutation checks confirmed that restoring block-based split produces `11 / 0`, and that an empty index list fails on Ray
- All repository pre-commit hooks passed

The branch is based on current `origin/main`; its commits use the repository GitHub noreply identity.


data_juicer/core/executor/ray_executor_partitioned.py
```diff
@@ -688,6 +701,12 @@ def _run_impl(self, load_data_np: Optional[PositiveInt] = None, skip_return=Fals
             # a no-op once the value is no longer "auto".
             self._resolve_max_concurrent_partitions(ops)
             self._configure_auto_partitioning(dataset, ops)
+        elif self.partition_size_cfg is not None:
+            self.num_partitions = self._partition_count_from_size(dataset, self.partition_size_cfg)
```
fengrui-z, line 715: These lines derive `num_partitions` from `partition.size`, but `_split_dataset_deterministic()` later calls `Dataset.split()` at lines 1640–1644. Ray distributes existing blocks among the requested splits without subdividing coarse blocks. For example, a single block containing 1,000 rows with `partition.size=100` can produce one non-empty partition and nine empty partitions, rather than approximately ten 100-row partitions. This makes the new setting ineffective and can undermine parallelism and checkpoint granularity. Please repartition first, or retain `total_samples` and use `split_at_indices()` to create row-based boundaries. A single-block regression test would also cover this case.

fengrui-z (commented):
When manual mode with partition.size derives num_partitions == 1, split_indices is empty and the code calls split_at_indices([]). Ray 2.51 raises ValueError: indices must be at least of length 1. [Relevant code](https://github.com/datajuicer/data-juicer/blob/0250101b151c52be9b2e9571bda8c2d811f28ec9/data_juicer/core/executor/ray_executor_partitioned.py#L1641-L1647)
Please special-case a single partition by returning [dataset.data.materialize()], and add a test where dataset.count() < 1.5 * partition.size.

fengrui-z (approved):
LGTM

PR #1074, opened 2026-09-21
fix(analysis): refine overall stats columns by the first non-null value

Linxiushen:
## Problem

`OverallAnalysis.refine_single_column` (`data_juicer/analysis/overall_analysis.py`) decides how to treat an `object` column from its first element only:

```python
first = col[0]
if type(first) not in self.supported_object_types:   # {str, list}
    logger.warning("There is a column of stats with type [...], which is not supported to be analyzed for now.")
    return None
```

If sample 0 happens to be `None` the entire column is dropped from `overall.csv` / `overall.md`, however many valid values follow. That is not a contrived input: `LLMExtractMapper` writes `{k: None for k in self.output_schema}` into `Fields.meta` for an empty input text (and after its retries are exhausted), and a `null` text in the first JSONL row is common. End to end with the real op and the real `OverallAnalysis.analyze()`:

```
meta written by the real op for sample 0: {'__dj__topic': None}
meta column: [None, 'Paris', 'Tokyo', 'Lima'] | dtype object | non-null 3 / 4
WARNING ... a column of stats with type [<class 'NoneType'>], which is not supported to be analyzed for now.
overall.csv columns: ['text_len']      <- '__dj__topic' (3 valid values) is gone
```

This is the default `dj-analyze` path (`Analyzer.run()` → `OverallAnalysis(...).analyze()`). The assumption "there won't be mixed types in a column" dates from #106 (2023), before the LLM mappers that legitimately emit `None` existed.

## Change

Decide the type from the first **non-null** value (`col.dropna()`). A column with no valid value at all is still skipped — same behaviour as today — but with an accurate warning ("without any valid value, which is skipped for now") instead of the misleading "type [NoneType] not supported". Genuinely unsupported types (dict, …) keep the existing warning and are still skipped.

Differential over 7 scenarios (all strings / all lists / all numbers / unsupported dict / mixed str+list / empty strings / NaN numbers): the full `analyze()` output is identical before and after on every well-formed input.

## Tests

`tests/analysis/test_overall_analysis.py`: `test_first_row_is_null` (a `None` at row 0 no longer drops the column — fails on `main` with `'stats_str' not found in RangeIndex(...)`) and `test_all_null_column` (an all-`None` column is still skipped). The three existing tests still pass; the rest of `tests/analysis/` shows the same environment-only failures locally before and after.

Related: #1052 touched this function and was withdrawn after the reviewer asked for a reproduction that fails on current `main` and a test that actually exercises the new branch — both are included here (the repro above uses `LLMExtractMapper` with no model call). `black` / `isort` / `flake8` (pre-commit pins) pass on the source file.

_Written with Claude Code (AI-assisted) and submitted under the account owner's authorization._


cmgzn (approved):
LGTM, Thanks for your work

